PR #5, opened 2016-02-14
Invalidate nodes

lbodor:


oehm-smith:
Ok with this and will merge.  My understanding still a bit thin and so I will take some time now to write / modify unit tests.

PR #32, opened 2016-03-23
Change opengeo geodb used

oehm-smith:
* To one that is still available

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,8 +175,8 @@
         </dependency>
         <dependency>
             <groupId>org.opengeo</groupId>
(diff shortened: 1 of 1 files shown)


lbodor:
Let's avoid third-party snapshot dependencies, because they expose us to the risk of non-repeatable builds. Let's consider a different solution. We can downgrade to version 0.7 or, preferably, package in our own copy of 0.8.

oehm-smith:
Closing - Lazar will copy the version he has into a lib/ directory and make a system dependency on that in the pom.

PR #60, opened 2016-04-06
Change to use Postgresql

simondobner:


oehm-smith:
It might be good to keep the Oracle information as well.  I'm not sure if our direction is to completely replace Oracle or just offer a PSQL alternative.  There is still data in the Oracle database that may be easier to extract via this application.

lbodor:
The old oracle jdbc connection string is still there in method `externalOracleDataSource.`

PR #109, opened 2016-04-22
Simplify test dependencies

oehm-smith:
I had problems when trying to integrate the email notifications code in as the tests pulled in every artifact via GeodesyServiceTestConfig via GeodesyServiceConfig.  The trouble with this is that every class, whether wanted or not and if relevant to the test or not are included and wired up.  This was seen to be a real issue in one fix where the test Asserted some number of records existed.  But that Service class (NodeService was one) wasn't explicitly mentioned in the test (ie. Autowired).  It was by virtue of being in the package that it was instantiated so that when the respective Repository.count() was called (eg. NodeRepository.count()) the value was what was expected (Asserted).

I have removed the use of GeodesyServiceTestConfig / GeodesyServiceConfig from all tests now and replaced with explicit X.class in the \@Configuration header.

This also removed the AspectJ weaving that was in GeodesyServiceConfig and I replaced the only use of this with different code.  That was in SiteLogSopacReader and its super-class SiteLogReader.

Now all the tests run in Eclipse also 🎱   And pass 👍

### DIFF
--- a/src/main/java/au/gov/ga/geodesy/domain/service/CorsSiteService.java
+++ b/src/main/java/au/gov/ga/geodesy/domain/service/CorsSiteService.java
@@ -41,7 +41,7 @@ import au.gov.ga.geodesy.domain.model.sitelog.EquipmentLogItem;
 import au.gov.ga.geodesy.domain.model.sitelog.SiteLog;
 import au.gov.ga.geodesy.domain.model.sitelog.SiteIdentification;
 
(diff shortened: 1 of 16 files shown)


oehm-smith:
@lbodor I've squashed the commits.

oehm-smith:
Rebased

lbodor:
Superseded by #133.

PR #111, opened 2016-04-28
Trial Isomorphism as abstraction for mappers

lbodor:


src/main/java/au/gov/ga/geodesy/support/mapper/orika/GnssReceiverOrikaMapper.java
```diff
@@ -41,11 +44,11 @@ public GnssReceiverOrikaMapper() {
         mapper = mapperFactory.getMapperFacade();
     }
 
-    public GnssReceiverLogItem mapFromDto(GnssReceiverType receiver) {
-        return mapper.map(receiver, GnssReceiverLogItem.class);
+    public Function<GnssReceiverType, GnssReceiverLogItem> to() {
+        return receiver -> mapper.map(receiver, GnssReceiverLogItem.class);
```
heyagosper: admit a lot of the java 8 syntax is new to me. could you explain where the receiver variable is defined/initialised?

src/main/java/au/gov/ga/geodesy/support/mapper/orika/GnssReceiverOrikaMapper.java
```diff
@@ -41,11 +44,11 @@ public GnssReceiverOrikaMapper() {
...
     }
 
-    public GnssReceiverType mapToDto(GnssReceiverLogItem logItem) {
-        return mapper.map(logItem, GnssReceiverType.class);
+    public Function<GnssReceiverLogItem, GnssReceiverType> from() {
+        return receiverLogItem -> mapper.map(receiverLogItem, GnssReceiverType.class);
```
heyagosper: same comment as above with receiverLogItem

src/main/java/au/gov/ga/geodesy/support/mapper/orika/GnssReceiverOrikaMapper.java
lbodor: It's a lambda expression, i.e., an anonymous function or a function value. It has one parameter called receiver, whose type is inferred at compile-time. Lambda expressions have existed in many languages for well over 20 years, and they have finally been added to Java.

PR #112, opened 2016-04-29
Accept multiple date patterns via FastDateFormat

lbodor:


src/main/java/au/gov/ga/geodesy/support/mapper/orika/DateToTimePositionConverter.java
```diff
@@ -13,27 +17,47 @@
...
+                        return empty;
+                    }
+                }
+            )
+            .filter(Optional::isPresent).map(Optional::get).findFirst();
+        if (date.isPresent()) {
+            return date.get();
+        } else {
+            String attemptedPatterns = StringUtils.join(timePositionPatterns, ", ");
+            throw new RuntimeException("Failed to parse " + dateString + " using the following patterns: " + attemptedPatterns);
```
heyagosper: as we discussed about lambdas earlier. I always feel like all you really get is more concise syntax, but at the expense of self-documenting code and enormous difficulty debugging. which means it's more important than ever to have documentation comments explaining what is going on.

PR #119, opened 2016-05-10
#113 Update since now using CountryCode CodeLists

oehm-smith:
* Due to Dozer working with super types, I couldn't have a CountryCode to/from String mapper.  Instead there is a Factory in the CodeLists to / from String mapper that selects that specific super types (either CountryCode or default (everything else))


oehm-smith:
Conflicts resolved

PR #125, opened 2016-05-10
Site log

lbodor:


src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/DateToTimePositionConverter.java
```diff
@@ -24,6 +24,7 @@
     private static final String[] timePositionPatterns = {
```
oehm-smith: I like your convertFrom() - much cleaner using streams than what I did.

oehm-smith:
This is all very neat and succinct with your use of Functions.  Fantastic work (I'm learning something).

PR #131, opened 2016-05-13
Use Orika's default compiler strategy

lbodor:


oehm-smith:
Uggh sorry for the formatting changes - getting used to IntelliJ way of things!

PR #163, opened 2016-05-26
#4 - Build Number

oehm-smith:
* Add timeFormated time stamp to Manifest
* Add git-branch to Manifest
* Removed project-version from Manifest as same
information available in Implementation-Version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -402,6 +402,17 @@
...
                     </execution>
+                    <execution>
+                        <id>timestamp-property</id>
+                        <goals>
+                            <goal>timestamp-property</goal>
+                        </goals>
+                        <configuration>
+                            <name>current.time</name>
+                            <pattern>yyyy-MM-dd HH:mm:ss Z</pattern>
+                            <timeZone>GMT+0</timeZone>
```
lbodor: Could we please use our local timezone here?

oehm-smith: For the record (this is merged after we discussed the issue), i could find no way through a maven plugin to use 'local timezone'.  What i did do is change the ManifestService to convert the time from GMT to the local timezone.  So unfortunately the GMT time is still in the manifest but the service will display the correct time.

PR #165, opened 2016-06-02
change date types to new date/time API

heyagosper:


src/main/java/au/gov/ga/geodesy/domain/model/converter/InstantConverter.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.sql.Timestamp;
+import java.time.Instant;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+
+/**
+ * Converts Instant objects to java.sql.Timestamp values that can be persisted in the database.
+ * Converts java.sql.Timestamp to Instant values that can be used in DTOs
```
lbodor: Not DTOs, but domain model entities.

src/main/java/au/gov/ga/geodesy/domain/model/converter/InstantConverter.java
```diff
@@ -0,0 +1,25 @@
+package au.gov.ga.geodesy.domain.model.converter;
```
lbodor: JPA converters are not part of the domain model, they are persistence mechanism utilities, so they belong in `au.gov.ga.geodesy.support.persistence.jpa`.

src/main/java/au/gov/ga/geodesy/domain/service/WeeklySolutionService.java
```diff
@@ -38,18 +39,16 @@ public void uploadSolution(String sinexFileName) {
...
+            Instant firstObservation = parseGpsEpoch(tokenizer.nextToken());
 
             @SuppressWarnings("unused")
-            Date lastObservation = parseGpsEpoch(tokenizer.nextToken());
+            Instant lastObservation = parseGpsEpoch(tokenizer.nextToken());
 
-            Calendar epoch = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-            epoch.setTime(firstObservation);
-            epoch.add(Calendar.DAY_OF_WEEK, 3);
+            firstObservation.plus(3, DAYS);
```
lbodor: This is a bug. Instants are immutable, so you are throwing away the new value. Also, even if you could modify firstObservation, it would stop being the time of the first observation. Keep the term "epoch" for this value, it the business domain term for the time of the weekly solution.

lbodor:
Please rebase to resolve merge conflicts.

PR #171, opened 2016-06-07
Reuse code in unit tests

heyagosper:


lbodor:
Rebase against master please. Consider that the order of arguments to assertEquals is significant with respect to actual value and expected value. Won't testMapping in the new super class give wrong error messages?

src/main/java/au/gov/ga/geodesy/support/utils/MappingDirection.java
```diff
@@ -0,0 +1,10 @@
+package au.gov.ga.geodesy.support.utils;
```
oehm-smith: This is only needed in Test

PR #176, opened 2016-06-07
Created a new mapper for FormInformation type

frankfu03:
Please review FormInformation mapper.


src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/FormInformationMapper.java
```diff
@@ -0,0 +1,44 @@
...
+ * FormInformation site log entity.
+ */
+public class FormInformationMapper implements Iso<FormInformationType, FormInformation> {
+
+    private MapperFactory mapperFactory = new DefaultMapperFactory.Builder().build();
+    private MapperFacade mapper;
+
+    public FormInformationMapper() {
+        mapperFactory.classMap(FormInformationType.class, FormInformation.class)
+                .fieldMap("datePrepared", "datePrepared").add()
```
lbodor: Is this line necessary? It looks like `byDefault` alone should do the job.

PR #179, opened 2016-06-08
Bump spring

lbodor:


pom.xml
```diff
@@ -172,7 +172,7 @@
         <dependency>
             <groupId>org.springframework.data</groupId>
             <artifactId>spring-data-jpa</artifactId>
-            <version>1.8.0.RELEASE</version>
```
oehm-smith: One advantage of SpringBoot is that it manages these transitive dependencies for you.  So you don't need to specify versions.  But it might be a mechanism independant of Spring Boot which we could employ.  I'll see if I can find it.

oehm-smith: Yep it looks like its via Spring Boot - http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#howto-customize-dependency-versions.  I'm still trying to find more info.

PR #181, opened 2016-06-08
Return location header when uploading site logs

lbodor:


src/test/java/au/gov/ga/geodesy/port/adapter/rest/UploadAliceIgsSiteLogRestTest.java
```diff
@@ -21,8 +24,14 @@
     @Rollback(false)
     public void uploadALIC() throws Exception {
         String content = FileUtils.readFileToString(TestResources.sopacSiteLog("ALIC"), Charset.defaultCharset());
-        mvc.perform(post("/siteLog/sopac/upload").contentType(MediaType.APPLICATION_XML).content(content))
-            .andExpect(status().isCreated());
+        MvcResult result = mvc.perform(post("/siteLog/sopac/upload").contentType(MediaType.APPLICATION_XML).content(content))
+            .andExpect(status().isCreated())
+            .andReturn();
+
+        String location = (String) result.getResponse().getHeaderValue("location");
```
oehm-smith: As per the team meeting this morning, we discussed:
* Get Sitelog from location
* assert that the object and the object.id is not null

PR #192, opened 2016-06-14
Rename equipment resource path element

lbodor:
This pull request explicitly sets the exported resource name to "equipment". By default, spring-data-rest sees the equipment entity class, and generates the "equipments" resource path.


oehm-smith:
Did you try:

    \@Repository("equipment")
    public class EquipmentsRepository .... {

lbodor:
I tried it just now, and it doesn't seem to work.

oehm-smith:
Ok no worries I just thought that was it's purpose.  I'll merge what you had.

PR #202, opened 2016-06-14
Fix problem with incomplete EffectiveDates in Dozer

oehm-smith:
* Fixed the case where the EffectiveDate from the data is null.
Previously 1/1/1970 TimeInstantType being returned.
Now returns TimePeriodType with empty lists for begin and end.
* Also fixed minor date tests

### DIFF
--- a/src/main/java/au/gov/ga/geodesy/support/mapper/dozer/converter/TimePrimitivePropertyTypeStringConverter.java
+++ b/src/main/java/au/gov/ga/geodesy/support/mapper/dozer/converter/TimePrimitivePropertyTypeStringConverter.java
@@ -23,8 +23,10 @@ public class TimePrimitivePropertyTypeStringConverter implements CustomConverter
             return null;
         }
         if (source instanceof String) {
+            TimePrimitivePropertyType castDestination = (TimePrimitivePropertyType) destination;
+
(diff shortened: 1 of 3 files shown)


lbodor:
Waiting on tests.

oehm-smith:
I've run UploadAliceGeodesyMLSiteLogRestTest with the newly generated files and there are no problems.  

Further to our discussion about representing no-date with an empty element, I add to my support for this that it forces a type.  `TimePositionType` can be an `TimeInstantType` or a `TimePeriodType`.  EffectiveDates have a begin and end so latter is preferred.  The XML for empty to/from dates looks like:

    <gml:validTime>
        <gml:TimePeriod gml:id="TimePeriodType_0">
            <gml:beginPosition></gml:beginPosition>
            <gml:endPosition></gml:endPosition>
        </gml:TimePeriod>
    </gml:validTime>

oehm-smith:
I've made the date changes as discussed and simplified by moving work to a separate branch and PR (yet to come).  The dates now work as described in the commit - https://github.com/GeoscienceAustralia/geodesy-domain-model/pull/202/commits/16a8809e2a2deae8ab1a82250ca02f78b26320b7

PR #204, opened 2016-06-15
Use receiverType field instead of igsModelCode

lbodor:


src/test/java/au/gov/ga/geodesy/port/adapter/rest/UploadAliceGeodesyMLSiteLogRestTest.java
```diff
@@ -1,6 +1,7 @@
 package au.gov.ga.geodesy.port.adapter.rest;
 
 import static io.restassured.module.mockmvc.RestAssuredMockMvc.given;
+import static org.hamcrest.Matchers.*;
```
oehm-smith: Its a better code style to be explicit about the imports used.

oehm-smith:
When did the field change from igsModelCode to ReceiverType?  I can't see it in the history of `Equipment.xsd`.

src/test/java/au/gov/ga/geodesy/port/adapter/rest/UploadAliceGeodesyMLSiteLogRestTest.java
lbodor: Generally, yes. However, modern attempts at using Java as a DSL host language are testing that convention. See: https://github.com/rest-assured/rest-assured/wiki/Usage#static-imports and http://www.javaworld.com/article/2074566/core-java/are-static-imports-becoming-increasingly-accepted-in-java-.html. Why not adopt a horses-for-courses approach and go with the flow.

lbodor:
It didn't change. It uses both igsModelCode and receiverType. I switched from one to the other. See https://github.com/GeoscienceAustralia/GeodesyML/issues/137.

src/test/java/au/gov/ga/geodesy/port/adapter/rest/UploadAliceGeodesyMLSiteLogRestTest.java
oehm-smith: ""ain't" is still generally frowned upon but is perhaps less frowned upon than it used to be"

I have no problems with static imports, just by `*` importing them.

Unless you have far too many to count, I think expanding the imports to be explicit is a better way to go.  I like to know the context of Asserts and the like.  For example - Hamcrest, Junit and TestNG (plus many more) implements Assert.  Or more importantly for me - I've seen the use of a method in someones code at code review time.  Where does it come from?  I like the context.  Your IDE should be able to resolve the imports.

I think explicitly resolving them adds a lot of value.  Not resolving them adds no value except you don't need to do the keystrokes to get your IDE to resolve them.  Unless there are too many and it then requires you to wade through all the imports.  But still you will only get imports for what you use.

On a slightly different note, I personally prefer to use non-static imports so that they are easier to resolve at the time of writing.  I'm actually not sure what happens in IDEA, but in Eclipse when you type `assert<ctrl-space>` nothing is shown in the list.  So you have to go and manually insert

    import static org.junit.Assert.*;

Which may require looking it up first.

Instead what I do is write `Assert.assert<ctrl-space>` to be shown `assertEquals(...) etc` AND let it `import org.junit.Assert` (after giving me a list to choose from).

So please, let us resolve static imports.

lbodor: Yes, well argued, I misunderstood what you meant at first. There is nothing to be gained by not listing the identifiers imported. I'll fix it.

PR #212, opened 2016-06-17
Add MultipathSource mapper

heyagosper:
Mapper and tests
Also added new mapper to SiteLogmapper and tests.


heyagosper:
@oehm-smith  could you have a look, I started using hamcrest (it was already in the pom dependencies)

heyagosper:
I've assigned you all just to bring to your attention that this mapper contains example of the hamcrest asserts and also that the SiteLogMapperTest is also using it now

PR #219, opened 2016-06-21
Radio interference mapper 2

oehm-smith:

PR #220, opened 2016-06-21
Some sonar corrections

oehm-smith:
Corrections for sonar feedback when working on radio interference mapper.  But wanted to separate the work as it isn't all related to radio interference mapper.  However there is some overlap so ...

DO NOT DO THIS BEFORE the radio interference mapper PR (https://github.com/GeoscienceAustralia/geodesy-domain-model/pull/219).

**FORGET THIS PR NOW (but not closing so can refer to it).  Due to merging this into #219 previously and because I have since rebased #219, there will be conflicts and resolving them won't be much fun.  It will be easier to do it all again.**  I will close this after that work is done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -360,6 +360,12 @@
             <artifactId>spring-context-support</artifactId>
             <version>${spring.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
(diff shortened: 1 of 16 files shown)


oehm-smith:
Closing and will redo the work.

PR #221, opened 2016-06-22
GEOD-21 Create GnssAntennas mapper

oehm-smith:


src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
```diff
@@ -0,0 +1,50 @@
...
+import ma.glasnost.orika.MapperFactory;
+import ma.glasnost.orika.converter.ConverterFactory;
+import ma.glasnost.orika.impl.DefaultMapperFactory;
+import ma.glasnost.orika.metadata.TypeFactory;
+import net.opengis.gml.v_3_2_1.CodeType;
+
+/**
+ * Reversible mapping between GeodesyML GnssAntennas DTO and
+ * RadioInterference site log entity.
+ */
```
lbodor: Incorrect comment.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
```diff
@@ -0,0 +1,50 @@
...
+                .register();
+
+
+        ConverterFactory converters = mapperFactory.getConverterFactory();
+        converters.registerConverter("codeWithAuthorityTypeConverter", new StringToCodeWithAuthorityTypeConverter("eGeodesy/type") {});
+        converters.registerConverter(new StringToCodeTypeConverter("eGeodesy/type") {});
+        converters.registerConverter("codeListValueTypeConverter", new StringToCodeListValueConverter<IgsReceiverModelCodeType>(
+                "https://igscb.jpl.nasa.gov/igscb/station/general/rcvr_ant.tab",
+                "http://xml.gov.au/icsm/geodesyml/codelists/antenna-receiver-codelists.xml#GeodesyML_GNSSReceiverTypeCode"
+        ));
```
lbodor: Incorrect code list value

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/SiteLogMapper.java
```diff
@@ -44,18 +46,19 @@
...
+                .fieldMap("pressureSensors", "pressureSensors").converter("pressureSensors").add()
+                .fieldMap("temperatureSensors", "temperatureSensors").converter("temperatureSensors").add()
+                .fieldMap("waterVaporSensors", "waterVaporSensors").converter("waterVaporSensors").add()
+                .fieldMap("otherInstrumentations", "otherInstrumentationLogItem").converter("otherInstrumentations").add()
+                .fieldMap("signalObstructionsSet", "signalObstructionLogItems").converter("signalObstructionsSet").add()
+                .fieldMap("multipathSourcesSet", "multipathSourceLogItems").converter("multipathSourcesSet").add()
+                .fieldMap("gnssAntennas", "gnssAntennas").converter("gnssAntennas").add()
             /* .byDefault() */
-            .register();
+                .register();
```
lbodor: Please try not to change at the same time the formatting and the content of the files you edit.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
```diff
@@ -0,0 +1,50 @@
...
+
+        mapperFactory.classMap(GnssAntennaLogItem.class, GnssAntennaType.class)
+                .fieldMap("type", "antennaType").converter("codeListValueTypeConverter").add()
+                .fieldMap("antennaRadomeType", "antennaRadomeType").converter("codeWithAuthorityTypeConverter").add()
+                .byDefault()
+                .register();
+
+
+        ConverterFactory converters = mapperFactory.getConverterFactory();
+        converters.registerConverter("codeWithAuthorityTypeConverter", new StringToCodeWithAuthorityTypeConverter("eGeodesy/type") {});
```
lbodor: Raw type, add type argument.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/StringToCodeWithAuthorityTypeConverter.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public String convertFrom(T code, Type<String> targetType, MappingContext ctx) {
+        return code.getValue();
+    }
+
+    public T convertTo(String str, Type<T> targetType, MappingContext ctx) {
+        T code = null;
+        try {
+            code = targetType.getRawType().newInstance();
+        } catch (InstantiationException | IllegalAccessException e) {
```
lbodor: I think that landing in this catch block is an indication of a bug that we need to know about. It is better to throw a subclass of runtime exception, rather than recover by returning null.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
```diff
@@ -0,0 +1,50 @@
...
+        converters.registerConverter(new StringToCodeTypeConverter("eGeodesy/type") {});
```
lbodor: Instead of using "eGeodesy/type" everywhere, choose a name that represents the code list, as in "eGeodesy/antennaRadomeType".

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/StringToCodeWithAuthorityTypeConverter.java
```diff
@@ -0,0 +1,37 @@
...
+            logger.error("convertTo error: ", e);
+        }
+        if (code != null) {
+            code.setValue(str);
+
+            code.setCodeSpace(codeSpace);
```
lbodor: Use the builder pattern if available.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
heyagosper: @lbodor the existing mappers do it this way, updating the others might be a good thing to add to friday squash

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/SiteLogMapper.java
oehm-smith: That means we have to format everything by hand, which defeats the purpose of having code formatting and style.  And increases the development time.  It also creates 'formatting' wars.

I think we should stick to a format and all use it everytime.  I'll change this here but moving forward I'd like to see this.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/StringToCodeWithAuthorityTypeConverter.java
oehm-smith: No none is 👎

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/SiteLogMapper.java
lbodor: I think you may have misunderstood me. We don't have to format everything by hand. I agree with you; we should use the formatter consistently. I only meant to say that we should try to separate the commits that change the formatting from the commits that change the content, but it's not that important.

More importantly, in this case, I suggest one level of indentation, rather than two. It would reduce the overall level of indentation, and not stress the breaking of expressions over lines, which is more and more common in Java.

heyagosper: as a side note, I'm annoyed by intelliJ indenting these chained method calls, I have to keep deleting spaces, I don't know how to change it.

oehm-smith:
Good to re-review now.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
```diff
@@ -0,0 +1,49 @@
...
+
+        mapperFactory.classMap(GnssAntennaLogItem.class, GnssAntennaType.class)
+                .fieldMap("type", "antennaType").converter("codeListValueTypeConverter").add()
+                .fieldMap("antennaRadomeType", "antennaRadomeType").converter("codeWithAuthorityTypeConverter").add()
+                .byDefault()
+                .register();
+
+
+        ConverterFactory converters = mapperFactory.getConverterFactory();
+        converters.registerConverter(new StringToCodeTypeConverter("eGeodesy/antennaRadomeType") {
```
lbodor: Is this converter needed? It's argument is specific to a field, but it isn't applied specifically. Further down there is another "eGeodesy/antennaRadomeType".

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/StringToCodeWithAuthorityTypeConverter.java
```diff
@@ -0,0 +1,37 @@
...
+        T code = null;
+        try {
+            code = targetType.getRawType().newInstance();
+        } catch (InstantiationException | IllegalAccessException e) {
+            throw new GeodesyRuntimeException("Error with convert: ", e);
+        }
+        if (code != null) {
+            code.setValue(str);
+            code.setCodeSpace(codeSpace);
+        } else {
```
lbodor: Unreachable code, class.newInstance does not return null ( https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/java/lang/Class.java#L16)

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
oehm-smith: Yes all converters used are the minimum viable set.

oehm-smith:
Pushed changes that addressed feedback.

src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/GnssAntennaMapper.java
```diff
@@ -0,0 +1,51 @@
...
+import au.gov.xml.icsm.geodesyml.v_0_3.GnssAntennaType;
+import au.gov.xml.icsm.geodesyml.v_0_3.IgsAntennaModelCodeType;
+import au.gov.xml.icsm.geodesyml.v_0_3.IgsRadomeModelCodeType;
+import ma.glasnost.orika.MapperFacade;
+import ma.glasnost.orika.MapperFactory;
+import ma.glasnost.orika.converter.ConverterFactory;
+import ma.glasnost.orika.impl.DefaultMapperFactory;
+
+/**
+ * Reversible mapping between GeodesyML GnssAntennas DTO and GnssAntennaLogItem site log entity.
```
heyagosper: not quite right comment - should refer to GnssAntennaType

PR #222, opened 2016-06-22
Add localEpisodicEvents Mapper and tests

heyagosper:
Required a database change, added to the liquibase changelog

please nobody merge this until Simon has had a chance to review and test it


heyagosper:
I also haven't added this mapper to the site log mapper yet so I'll add a new commit for that

src/main/java/au/gov/ga/geodesy/domain/model/sitelog/LocalEpisodicEventLogItem.java
```diff
@@ -22,7 +22,7 @@
 
     @Size(max = 256)
     @Column(name = "EVENT_DATE", length = 256)
```
lbodor: This column mapping shouldn't be here. The column mappings are provided by EffectiveDates class. I'm not sure if what you get in the database with this mapping present. Traditionally, when you nest another classes columns in your table, the annotation to use is @Embedded, see Setup.java, for instance.

heyagosper:
superceded by #224

PR #225, opened 2016-06-22
Query DSL

lbodor:
https://spring.io/blog/2011/04/26/advanced-spring-data-jpa-specifications-and-querydsl/
http://www.querydsl.com/


oehm-smith:
It unfortunately breaks IntelliJ - the AJC weaving is at fault I believe.  Looking at it now.

oehm-smith:
That's pretty neat.  Nice.

PR #228, opened 2016-06-24
Fix bug in LocalEpisodicEvents

oehm-smith:
* Specifically with EffectiveDates
* And test data used for LocalEpisodicEvents

### DIFF
--- a/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/EffectiveDates.java
+++ b/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/EffectiveDates.java
@@ -1,9 +1,12 @@
 package au.gov.ga.geodesy.domain.model.sitelog;
 
-import java.time.Instant;
+import afu.org.apache.commons.lang3.builder.EqualsBuilder;
+import afu.org.apache.commons.lang3.builder.HashCodeBuilder;
+import afu.org.apache.commons.lang3.builder.ToStringBuilder;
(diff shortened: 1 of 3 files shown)


src/main/java/au/gov/ga/geodesy/domain/model/sitelog/EffectiveDates.java
```diff
@@ -56,21 +56,28 @@ public boolean equals(Object x) {
...
         if (to == null) {
-            return other.getTo() == null && other.getFrom().equals(from);
+            if (other.getTo() != null) {
+                return false;
+            }
+        } else if (!to.equals(other.getTo())) {
+            return false;
         }
-        return from.equals(other.getFrom()) && to.equals(other.getTo());
+        return true;
```
lbodor: Let's generally use EqualsBuilder from apache commons to implement our equals methods.

heyagosper: yes. it's easy to forget, especially when you're used to writing/modifying them by hand. should we make a friday task for someone to go and change the existing ones? we might steamroll some bugs that we hadn't tested for.

oehm-smith:
Updated to use Apache commons lang3 builders.  Also tried to use `CompareToBuilder` for the `compareTo` but it didn't work and I decided not to put any more effort into it.

PR #229, opened 2016-06-24
Search for setups using Query DSL

lbodor:


src/test/java/au/gov/ga/geodesy/port/adapter/rest/SetupEndpointTest.java
```diff
@@ -38,13 +40,19 @@ public void testFindCurrentByFourCharacterId() throws Exception {
     public void testFindByFourCharacterIdAndDate() throws Exception {
         mvc.perform(get("/setups/search/findByFourCharacterId?id=ALIC&effectiveFrom=2011-12-12&effectiveTo=2014-12-12&timeFormat=uuuu-MM-dd"))
             .andExpect(status().isOk())
-            .andExpect(jsonPath("$.page.totalElements").value(3));
+            .andExpect(jsonPath("$.page.totalElements").value(3))
```
oehm-smith: You could use a value like you did in the test below.  Or change that one (be consistent).

PR #231, opened 2016-06-27
#214 Guard against NPE

lbodor:


src/main/java/au/gov/ga/geodesy/domain/model/CorsSite.java
```diff
@@ -18,63 +21,58 @@
...
     @Column(name = "BEDROCK_TYPE")
-    private String bedrockType;
+    private @MonotonicNonNull String bedrockType;
 
     @Column(name = "BEDROCK_CONDITION")
-    private String bedrockCondition;
+    private @MonotonicNonNull String bedrockCondition;
 
-    @SuppressWarnings("unused") // hibernate needs the default constructor
+    @SuppressWarnings({"unused", "initialization.fields.uninitialized"}) // hibernate needs the default constructor
```
oehm-smith: Minor and irrelevant, but the comment applies to the constructor and not the annotation so would be better inside the body.  Plus means it will be less difficult to delete.

src/main/java/au/gov/ga/geodesy/domain/model/sitelog/SiteLog.java
```diff
@@ -34,111 +35,109 @@
...
     @Valid
     @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
     @JoinColumn(name = "SITE_ID", referencedColumnName = "ID")
-    protected Set<LocalEpisodicEventLogItem> localEpisodicEventLogItems;
+    protected Set<LocalEpisodicEventLogItem> localEpisodicEventLogItems = new HashSet<>();
 
     @Valid
     @Embedded
-    protected MoreInformation moreInformation;
+    protected MoreInformation moreInformation = new MoreInformation();
```
oehm-smith: Lets check this to see if this works ok after mapping.

lbodor: I've created a Jira issue for this (https://gaautobots.atlassian.net/browse/GEOD-48). The issue applies to all our Orika mappers. Let's merge this pull request and sort out the issue independently.

PR #247, opened 2016-07-04
Re-add integration tests

oehm-smith:
* Was previously deleted
* Re-added but Using a property based approach so there remains only one 'surefire-plugin'.


oehm-smith:
Updated during recent discussions. Please review.

PR #249, opened 2016-07-05
Implement true query-by-example for receivers

lbodor:


oehm-smith:
That is so much more easier to understand!

PR #267, opened 2016-07-12
Generate documentation site

lbodor:


oehm-smith:
Where do we see the documentation?  Could you put a wiki page up.

PR #328, opened 2016-11-02
Document patch

lbodor:


oehm-smith:
Python API call by Lazar at https://gist.github.com/lbodor/5b83c19ec4dd8d44415f1d23396f2fa3

PR #337, opened 2016-11-14
Map responsible parties

lbodor:


src/main/java/au/gov/ga/geodesy/domain/model/sitelog/SiteLog.java
```diff
@@ -424,31 +429,30 @@ public void setMoreInformation(MoreInformation value) {
     }
 
     /**
-     * @return the siteContact
+     * Return responsible parties associated with this site.
      */
-    public @Nullable SiteResponsibleParty getSiteContact() {
-        return siteContact;
+    public List<SiteResponsibleParty> getResponsibleParties() {
```
oehm-smith: Are the setters missing?

PR #351, opened 2016-12-05
Refactor into mutli-module repository

lbodor:
Merge parent, domain, and web into a maven multi-module mega repository.


heyagosper:
what does the gws- prefix stand for? it feels like this should be "geodesy-" or something a bit easier to scan

lbodor:
It stands for Geodesy Web Services: gws-parent, gws-core, and gws-webapp.

PR #355, opened 2016-12-12
Split tests

lbodor:


gws-core/pom.xml
```diff
@@ -589,22 +614,87 @@
...
+            </activation>
+            <properties>
+                <skip.unit.tests>false</skip.unit.tests>
+                <skip.integration.tests>true</skip.integration.tests>
+            </properties>
+        </profile>
+        <profile>
+            <id>integration-test</id>
+            <properties>
+                <skip.unit.tests>true</skip.unit.tests>
```
oehm-smith: You should run the UTs before the ITs

lbodor: I have done that now.
* `mvn test` runs unit tests
* `mvn verify` runs unit and integration tests
* `mvn verify -Pintegration-test` runs integration tests
* `mvn deploy` runs unit tests, integration tests, and deploys

:+1: Thank you for your feedback, we are in a better position now.

PR #356, opened 2016-12-14
Compose

lbodor:
Run

* `docker-compose up` to start the database and web server
* `ctrl-c` to shutdown
* `docker-compose rm` to delete the database and web server containers

### DIFF
--- a/gws-webapp/src/main/docker/webserver/entrypoint.sh
+++ b/gws-webapp/src/main/docker/webserver/entrypoint.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
+
+java_link="$(which java)"
+java=$(readlink -f "${java_link}")
+java_bin=$(dirname "${java}")
+JAVA_HOME=$(dirname "${java_bin}")
+export JAVA_HOME
+
+catalina_link="$(which catalina.sh)"
+catalina=$(readlink -f "${catalina_link}")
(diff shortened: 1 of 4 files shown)


gws-webapp/src/main/docker/webserver/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
+
+java_link="$(which java)"
+java=$(readlink -f "${java_link}")
+java_bin=$(dirname "${java}")
+JAVA_HOME=$(dirname "${java_bin}")
```
oehm-smith: Shouldn't it be:

````
JAVA_HOME=$(dirname "${java}")
java_bin=$(JAVA_HOME/bin)
```

And you don't seem to add `java_bin` to the path.

I'm guessing then that you don't actually want (or NEED) java and tomcat below in the PATH.  And you are using the `_bin` suffix to just say 'its the binary'.

gws-webapp/src/main/docker/webserver/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+export JAVA_HOME
+
+catalina_link="$(which catalina.sh)"
```
oehm-smith: is `catalina.sh` in the PATH?  Oh, I'm guessing that NIX does that.

PR #361, opened 2016-12-16
Bump maven docker plugin

lbodor:


gws-core/pom.xml
```diff
@@ -594,7 +594,7 @@
             <plugin>
               <groupId>io.fabric8</groupId>
               <artifactId>docker-maven-plugin</artifactId>
-              <version>0.16.9</version>
+              <version>0.18.1</version>
```
oehm-smith: How about making a property up the top.

PR #370, opened 2016-12-19
gws-core, remove unused maven profiles

lbodor:
These profiles are in gws-parent.


gws-core/pom.xml
```diff
@@ -814,83 +814,79 @@
...
-                        <configuration>
-                            <excludes>
-                                <exclude>**/*</exclude>
-                            </excludes>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
+        <!-- <profile> -->
```
oehm-smith: Dont want to just delete?

PR #381, opened 2017-01-24
add delete_date and delete_reason columns into sitelog tables

hjin01:
add two columns "delete_time" and "delete_reason" into sitelog_tables;
also add comments on two new columns


oehm-smith:
:+1:

PR #382, opened 2017-01-25
Removed redundant BigDecimal mapper

heyagosper:


lbodor:
:+1:

lbodor:
I think you don't even need a PR provided you're just deleting code. What could there be to approve?

PR #392, opened 2017-01-31
Configure OpenAM with test users and groups

lbodor:


oehm-smith:
:+1:  (yes I see its merged already!)

PR #397, opened 2017-02-03
Refactor to use correct element  name

heyagosper:


gws-core/src/main/java/au/gov/ga/geodesy/support/mapper/orika/geodesyml/SiteLogMapper.java
```diff
@@ -74,7 +72,7 @@ public SiteLogMapper() {
             .fieldMap("otherInstrumentations", "otherInstrumentationLogItem").converter("otherInstrumentations").add()
             .fieldMap("signalObstructionsSet", "signalObstructionLogItems").converter("signalObstructionsSet").add()
             .fieldMap("multipathSourcesSet", "multipathSourceLogItems").converter("multipathSourcesSet").add()
-            .fieldMap("localEpisodicEventsSet", "localEpisodicEventLogItems").converter("localEpisodicEventsSet").add()
+            .fieldMap("localEpisodicEffects", "localEpisodicEventLogItems").converter("localEpisodicEffects").add()
```
oehm-smith: Isn't there already localEpisodicEffects (hence why the plural is ...Sets)?

heyagosper: it is "Set" because it is wrong

PR #406, opened 2017-02-09
Configure OpenAM Oauth2 redirects

lbodor:
Add http://localhost:5555/login and http://localhost:5555.


oehm-smith:
:+1:  :pizza:

PR #409, opened 2017-02-10
Use notes element from RemarksGroup

heyagosper:


oehm-smith:
:+1: :1st_place_medal:

PR #413, opened 2017-03-03
Reconfigure OpenAM docker image

lbodor:
Users are user.a, user.b, and user.x with authorities edit-ade1, edit-ade2,
and superuser, and passwords gumby123A, gumby123B, and gumby123X,
respectively.

RSA key pair is the default OpenAM `test` keypair.


oehm-smith:
:+1:  :apple:

PR #420, opened 2017-03-15
Make docker cache each package installation

lbodor:


oehm-smith:
Wow what a small change for what would seem like a large difference.  Well done!  :+1:

PR #422, opened 2017-03-26
Add user registration

lbodor:
* Rest endpoint
* Data model
* SNS notification


oehm-smith:
:+1:  Looks good.  Do you want to attempt an ITest with AWS SNS?

lbodor:
Yes, in my next PR. It will be a system test.

PR #425, opened 2017-03-28
delete old test data and add new test data

heyagosper:


oehm-smith:
Eee gads!  I will assume the translate is correct.  And given that, :+1:   :fleur_de_lis:

PR #433, opened 2017-04-06
Narrow down AWS SDK dependency

lbodor:


oehm-smith:
:+1: :large_blue_diamond:

PR #441, opened 2017-04-12
Set up system test URLs for test environment

lbodor:


gws-system-test/src/test/resources/system-test.properties
```diff
@@ -13,6 +13,5 @@ oauthProviderUrl = http://localhost:8083/openam/oauth2
 # oauthProviderUrl = https://devgeodesy-openam.geodesy.ga.gov.au/openam/oauth2
 
 # test
-# webServicesUrl = https://test.geodesy.ga.gov.au
-
-
+# webServicesUrl = https://testgeodesy-webservices.geodesy.ga.gov.au
```
oehm-smith: Did you mean to leave them commented out?

lbodor: Yes, you choose the one environment against which you want to run systems tests.

PR #454, opened 2017-05-04
Stop downloading the latest war file

lbodor:
Now that AWS code-deploy scripts are in geodesy-web-services, we already
have the correct version to deploy.


frankfu03:
:+1:

PR #462, opened 2017-05-17
Respect the order of responsible parties

lbodor:
The order of responsible party types is defined in GeodesyML in `siteLog.xsd`.
The order of responsible party values is defined by the order in which
they appear in a given GeodesyML site log document.


oehm-smith:
:+1: apart from failing build

PR #465, opened 2017-05-29
Remove emailing support, AWS SNS is simpler

lbodor:


oehm-smith:
Do you need to replace with something to connect to SNS?

PR #469, opened 2017-05-30
Lookup SNS topics by stack, as well as event name

lbodor:


gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/sns/SnsNotificationAdapter.java
```diff
@@ -40,7 +41,12 @@ public SnsNotificationAdapter() {
         // An SNS topic ARN looks like this: arn:aws:sns:ap-southeast-2:094928090547:DevGeodesy-UserRegistrationReceived-K3F2UQVHG58F
         return sns.listTopics().getTopics().stream()
             .filter(topic -> {
-                return notification.getSubject().equals(topic.getTopicArn().split(":")[5].split("-")[1]);
+                String[] tmp = topic.getTopicArn().split(":")[5].split("-"); // { "DevGeodesy", "UserRegistrationReceived", "K3F2UQVHG58F }
+                String stackName = tmp[0];
+                String eventName = tmp[1];
+
+                return Aws.getStackName().map(s -> s.equals(stackName)).orElse(false)
+                    && notification.getSubject().equals(eventName);
```
oehm-smith: Could it be a problem with using a valid notification?  Perhaps you should log if no corresponding topic is found.

lbodor: Good idea. I'll send another pull request to log a message when no topic could be found.

PR #472, opened 2017-05-31
To find stack name also match on instance id

lbodor:


gws-core/src/main/java/au/gov/ga/geodesy/support/aws/Aws.java
```diff
@@ -17,14 +17,19 @@
...
-        return EC2MetadataUtils.getInstanceId() != null;
+        return getInstanceId() != null;
+    }
+
+    public static final String getInstanceId() {
+        return EC2MetadataUtils.getInstanceId();
     }
 
     public static Optional<String> getStackName() {
+        String instanceId = getInstanceId();
```
oehm-smith: Does the compiler not complain that you need to access this as a static?

gws-core/src/main/java/au/gov/ga/geodesy/support/aws/Aws.java
```diff
@@ -17,14 +17,19 @@
...
         if (inAmazon()) {
             DescribeTagsResult tags = ec2.describeTags();
             return tags.getTags().stream()
-                .filter(t -> t.getKey().equals("aws:cloudformation:stack-name"))
+                .filter(t -> instanceId.equals(t.getResourceId()) && "aws:cloudformation:stack-name".equals(t.getKey()))
```
oehm-smith: To me this feels like your code is tightly coupled to expectation that there is a key with that name.

lbodor: It is.

gws-core/src/main/java/au/gov/ga/geodesy/support/aws/Aws.java
lbodor: It doesn't.

PR #496, opened 2017-06-09
add exported configuration files

heyagosper:


heyagosper:
there are a couple of small changes in the files that I didn't expect and don't understand

lbodor:
You may have more changes in there than you indented to make. For example, I still see mention of attribute `title`, which I remember was just an experiment. One way to proceed would be to recreate your container from `next` branch, make just the changes that you want to make (as per the instructions you are writing), and re-export the configuration files.

PR #500, opened 2017-06-15
Release candidate

lbodor:


oehm-smith:
Previously approved so :+1:

PR #645, opened 2019-11-27
Create GNSS documents bucket using terraform

frankfu03:
The name of the S3 bucket ( gnss-documents-{env} ) is open for discussion. There is a bucked named ga-gnss-documentation in s3, which has no environment suffix.


aws/terraform/s3bucket.tf
```diff
@@ -0,0 +1,15 @@
+resource "aws_s3_bucket" "gnss-documents" {
```
lbodor: underscores here: gnss_documents

frankfu03: Ok, it is done.

frankfu03:
As suggested, I have rename the S3 bucket to "gnss-metadata-document-storage-${var.environment}".

frankfu03:
The PR is ready for review again.

frankfu03:
As suggested, I have rename s3bucket.tf to document_storage.tf.

PR #656, opened 2020-01-31
Associated document endpoint

frankfu03:


lbodor:
Shouldn't this branch be based on "associated-document-bucket-terraform", so that it can make use of the terraform code to create the bucket?

frankfu03:
Yes, on my PC, I have rebased this PR against in "associated-document-bucket-terraform" branch and successfully run the integration test.

The class AssociatedDocument.java was merged into master in previous PR.

gws-core/pom.xml
```diff
@@ -1027,5 +1038,63 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>run-localstack</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
```
lbodor: Did you get the idea to use profiles from geodesy-archive/maven-parent/pom.xml? Profies are used in geodesy-archive to provide a mechanism for each subcomponent (i.e., command/submit-rinex-file) to choose the containers it wants to run. In GWS that seems unnecessary, because there aren't any separate subcomponents, and it would be simpler to configure the localstack container in the same way that the postgres container is currently configured.

lbodor: With the addition localstack, executions with ids "start-db" and "stop-db" should be renamed to something like "start-containers" and "stop-containers", because they now apply to both postgres and localstack containers.

lbodor: Executions "start-site-db" are used to bring up the database container so that a DDL script can be generated and uploaded to the documentation site. This execution now also brings up localstack, which isn't needed during site generation. We need a mechanism to choose what containers to start based on the maven phase, i.e., postgres and localstack should run during the pre-integration-test phase, and postgres should rub during the pre-site phase.

lbodor: [This new feature](https://github.com/fabric8io/docker-maven-plugin/pull/1259) recently added to maven-docker-plugin should help select docker images based on maven phas

lbodor: I'll prepare a PR to address the above.

frankfu03: Thanks. The code/logic looks better after merged the PR #657.

aws/deploy-local.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+set -eE
```
lbodor: I like the `-E`, I suspect we should start adding it to our scripts. either all or just those that set traps. I think that in your case, the full "strict" bash preamble should be `set -eEuo pipefail`.

aws/deploy-local.sh
```diff
@@ -0,0 +1,54 @@
...
+terraform apply -auto-approve -input=false
+
+function onExit {
+    rm -f main_override.tf
+}
+
+onExit
+trap onExit ERR
+trap "onExit; exit 1" SIGINT
+popd
```
lbodor: no need for popd

aws/deploy-local.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+environment="local"
+export TF_VAR_region=$AWS_DEFAULT_REGION
+export TF_VAR_environment=$environment
+
+cd "$(dirname "$(realpath "$0")")"
+cd ../
+
+pushd aws/terraform
```
lbodor: You can just use `cd`, changing directory in a shell script is a local effect.

aws/deploy-local.sh
```diff
@@ -0,0 +1,54 @@
...
+  version = "~> 2.22.0"
+  s3_force_path_style = true
+  endpoints {
+    s3 = "http://localhost:4572"
+  }
+}
+EOF
+
+terraform init
+terraform workspace select $environment || terraform workspace new $environment
```
lbodor: Are you using shellcheck?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+    public AssociatedDocumentEndpoint() {
+        AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext();
+        context.register(AmazonS3ContextConfiguration.class);
+        context.refresh();
+        this.s3Client = (AmazonS3) context.getBean("s3Client");
+        this.bucketName = (String) context.getBean("bucketName");
+    }
+
+    @PostMapping(
+        value = "/upload",
```
lbodor: Could we just pust to /associatedDocuments instead, and not use /upload. See discussion about APIs in slack.

aws/deploy-local.sh
frankfu03: OK, changed to `cd`.

aws/deploy-local.sh
frankfu03: Yeah, changed to` cd`.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+            objectMetadata.setContentType(file.getContentType());
+
+            PutObjectRequest putObjectRequest = new PutObjectRequest(this.bucketName,
+                    documentName, file.getInputStream(), objectMetadata)
+                    .withCannedAcl(CannedAccessControlList.PublicRead);
+            s3Client.putObject(putObjectRequest);
+            String objectUrl = s3Client.getUrl(this.bucketName, documentName).toExternalForm();
+            log.info("Uploaded " + documentName + " to S3 bucket: " + objectUrl);
+            return new ResponseEntity<String>(objectUrl, HttpStatus.OK);
+        } catch (Exception e) {
```
lbodor: Try to register an @ExceptionHandler in this controller to avoid catching exceptions here. The exception handler declared outside the body of this method will handle more situtations.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+        try {
+            if (doesBucketContainObject(this.bucketName, documentName)) {
+                DeleteObjectRequest deleteObjectRequest = new DeleteObjectRequest(this.bucketName, documentName);
+                s3Client.deleteObject(deleteObjectRequest);
+                log.info("Delete " + documentName + " from S3 bucket " + this.bucketName + ": done");
+            }
+            return new ResponseEntity<String>(HttpStatus.OK);
+        } catch (Exception e) {
+            log.error(e.getMessage(), e);
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
```
lbodor: Can't say "not found" in face of any exception.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @DeleteMapping("/{name}")
+    public ResponseEntity<String> remove(@PathVariable("name") String documentName) {
+        try {
+            if (doesBucketContainObject(this.bucketName, documentName)) {
+                DeleteObjectRequest deleteObjectRequest = new DeleteObjectRequest(this.bucketName, documentName);
+                s3Client.deleteObject(deleteObjectRequest);
+                log.info("Delete " + documentName + " from S3 bucket " + this.bucketName + ": done");
+            }
+            return new ResponseEntity<String>(HttpStatus.OK);
```
lbodor: A better response code for deletes is 204, not 200.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @DeleteMapping("/{name}")
+    public ResponseEntity<String> remove(@PathVariable("name") String documentName) {
+        try {
+            if (doesBucketContainObject(this.bucketName, documentName)) {
+                DeleteObjectRequest deleteObjectRequest = new DeleteObjectRequest(this.bucketName, documentName);
+                s3Client.deleteObject(deleteObjectRequest);
+                log.info("Delete " + documentName + " from S3 bucket " + this.bucketName + ": done");
+            }
```
lbodor: It may be more appropriate to return 404 when the object does not exist. Idempotency of the delete method refers to server-state, not the response returned. What is the general advice?

aws/deploy-local.sh
frankfu03: No. I installed "Shell Script"  for IntelliJ IDEA, RubyMine, WebStorm,

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+            log.error(e.getMessage(), e);
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
+        }
+    }
+
+    private boolean doesBucketContainObject(String bucketName, String objectName) {
+        try {
+            return s3Client.doesObjectExist(bucketName, objectName);
+        } catch (Exception e) {
+            return false;
```
lbodor: This could be a RuntimeException, like null pointer, which means you have a bug in your code, not that the object does not exist. The effect of above is to conceal the bug.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+        @Bean
+        public AmazonS3 s3Client() {
+            return AmazonS3ClientBuilder
+                .standard()
+                .withPathStyleAccessEnabled(true)
+                .withRegion(Regions.AP_SOUTHEAST_2)
+                .withCredentials(
+                    new AWSCredentialsProviderChain(
+                        new InstanceProfileCredentialsProvider(false),
+                        new ProfileCredentialsProvider("gnss-metadata"),
```
lbodor: This code runs in AWS, where I suspect there won't be any "gnss-metadata" credentials profile.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    public static class AmazonS3ContextConfiguration {
+        @Bean
+        public AmazonS3 s3Client() {
+            return AmazonS3ClientBuilder
+                .standard()
+                .withPathStyleAccessEnabled(true)
+                .withRegion(Regions.AP_SOUTHEAST_2)
```
lbodor: The region needs to be configurable at deployment time, not hard-coded.

gws-core/src/test/java/au/gov/ga/geodesy/support/spring/IntegrationTest.java
```diff
@@ -63,6 +64,10 @@ public void setUp() {
                 .uris().withScheme("https").withHost("gws.geodesy.ga.gov.au").withPort(443))
             .build();
         RestAssuredMockMvc.mockMvc(mvc);
+
+        // Fix SdkClientException: Unable to verify integrity of data download
+        // https://github.com/localstack/localstack/issues/869
```
lbodor: Add a TODO here so we can find this comment.

gws-core/src/test/java/au/gov/ga/geodesy/support/spring/IntegrationTestConfig.java
```diff
@@ -41,11 +49,29 @@
     @Value("${dbPassword}")
     private String dbPassword;
 
+    @Value("${gnss_metadata_document_bucket_name}")
```
lbodor: Mixed naming convention, the 3 lines above use camel case.

gws-core/src/test/java/au/gov/ga/geodesy/support/spring/IntegrationTestConfig.java
```diff
@@ -41,11 +49,29 @@
...
+    private String documentBucketName;
+
     @Bean
     public static PropertySourcesPlaceholderConfigurer propertySourcesPlaceholderConfigurer() {
         return new PropertySourcesPlaceholderConfigurer();
     }
 
+    @Bean
+    public String gnssMetadataDocumentBucketName() {
```
lbodor: Heads up, simple strings like this should be beans, but I don't have an example of an alternative outside of the unmerged branch "access-tokens" in gnss-data.

aws/deploy-local.sh
```diff
@@ -0,0 +1,54 @@
...
+if [[ $# -gt 0 ]]; then
+    echo "No argument excepted"
+    usage
+fi
+
+environment="local"
+export TF_VAR_region=$AWS_DEFAULT_REGION
+export TF_VAR_environment=$environment
+
+cd "$(dirname "$(realpath "$0")")"
```
lbodor: There is an inconsistency here between ${BASH_SOURCE[0]} and $0, what's the preference? Since the former is exists and is more complicated than the later, I suspect that there is probably some advantage to it. Is there? What is it?

aws/deploy-local.sh
frankfu03: Good, it is done for both deploy.sh and deploy-local.sh.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Changed to HttpStatus.NO_CONTENT (204).

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: In general it will be a 400 series error if the request is wrong somehow, and a 500 series error if something goes awry on the server. In most cases it would return 404 (NOT_FOUND).

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: It is hard to decide which status code is more appropriate. It could be status code of 4** or 5**.

lbodor: I think it's 500 for an unknown error, which is what `Exception` means. A NullPointerException is a subclass of Exception, and it indicates a bug, in which case it needs to be HTTP 500. Returning 404 would hide the bug.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Removing "/upload" from @PostMapping got the Status code 403 (Forbidden). There are no PathVariable or RequestParam, only a RequestPart in the POST URL. I guess we have to keep "/upload" in URL.

frankfu03: I found out that I need to make relevant changes in ResourceServerConfig class as well. It is now working after removed "/upload" from the POST URL.

lbodor: I don't think so. What is the cause of the HTTP 403 error? Somewhere there is a class that specifies permissions for all exposed endpoints.

lbodor: Yes, that's it.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: It was also hardcoded in Aws.java and SnsNotificationAdapter.java. I may fix this issue for relevant java classes in a separate PR.

gws-core/src/test/java/au/gov/ga/geodesy/support/spring/IntegrationTest.java
frankfu03: Done, added a "TODO ..." to the comment.

gws-core/src/test/java/au/gov/ga/geodesy/support/spring/IntegrationTestConfig.java
frankfu03: Changed to camel case "gnssMetadataDocumentBucketName" for this class and AssociatedDocumentEndpoint.java.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
lbodor: Yes, but your code is returning 200 when the object doesn't exist.

aws/deploy-local.sh
frankfu03: $0 is the path to the script. However, when a script is sourced, such as . a.sh, a.sh‘s $0 does not give a.sh but the caller’s name. While ${BASH_SOURCE[0]} can get the script’s own path, in sourced or directly executed Bash script.
So we may use ${BASH_SOURCE[0]} for all bash scripts.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: I have register an @ExceptionHandler in AssociatedDocumentEndpoint.

aws/deploy-local.sh
lbodor: Sounds good to me.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,134 @@
...
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
+        }
+    }
+
+    @ExceptionHandler(value = {AmazonServiceException.class, SdkClientException.class})
+    public ResponseEntity<String> handleSdkClientException(Exception e) {
+        log.error(e.getMessage(), e);
+        if (e instanceof AmazonServiceException) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(e.getMessage());
+        } else if (e instanceof AmazonServiceException) {
```
lbodor, line 103: unreachable code, did you mean SdkClientException?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
+        }
+    }
+
+    @ExceptionHandler(value = {AmazonServiceException.class, SdkClientException.class})
+    public ResponseEntity<String> handleSdkClientException(Exception e) {
+        log.error(e.getMessage(), e);
+        if (e instanceof AmazonServiceException) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(e.getMessage());
```
lbodor, line 102: How do we know that AmazonServiceException means that the client send a bad request to our API?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,134 @@
...
+        } else if (e instanceof AmazonServiceException) {
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(e.getMessage());
+        }
+        return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(e.getMessage());
```
lbodor, line 106: How do we know at this point that the client sent a bad request to our API?

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/EndpointsITest.java
```diff
@@ -28,6 +28,9 @@ public void hitAllEndpoints() throws Exception {
 
         Json.jq(rootResponse, "._links[].href | capture(\"https://[^/]*(?<url>[^{]*)\") | .url").stream()
             .map(JsonNode::asText)
-            .forEach(url -> given().get(url).then().statusCode(200));
+            .forEach(url -> {
+                System.out.println("~~~~~~~~~~~ URL=" + url);
```
lbodor, line 32: debugging code

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Yes, it could be a problem. Do you have any suggestion? The AWS profile "geodesy" for geodesy-operations seems work fine here.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Yes, you are right. I'll correct it.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpointITest.java
```diff
@@ -0,0 +1,46 @@
...
+
+
+public class AssociatedDocumentEndpointITest extends IntegrationTest {
+
+    private String documentName = "ALIC_ant_000_20191027T143000.jpg";
+    private String fileReference = "http://localhost:4572/gnss-metadata-document-storage-local/" + this.documentName;
+
+    @Test
+    @Rollback(false)
+    public void uploadSiteImageToS3Bucket() throws Exception {
```
lbodor, line 27: uploadSiteImage may be a better name, since the use of an s3 bucket is private to the implementation

lbodor, line 27: Similarly, it's not a site image at this point, but any document.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Sorry, the code should be:

```
        if (e instanceof SdkClientException) {
            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(e.getMessage());
        } else if (e instanceof AmazonServiceException) {
            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(e.getMessage());
        }
```

So, AmazonServiceException returns INTERNAL_SERVER_ERROR as it is any errors occurred in Amazon S3 while processing the request, while SdkClientException returns bad request as it is any errors are encountered in the client while making the request or handling the response.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/EndpointsITest.java
frankfu03: yep, will remove it.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpointITest.java
frankfu03: good.

frankfu03: Good. I'll change it to uploadDocument()

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Actually,  I am not know. This is for IOException. The document to be uploaded could be Not Acceptable, Not Found,  Unsupported Media Type and so on.

frankfu03: Maybe I should change it to return Not Found as it is IOException. What do you think?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
lbodor: Returning "bad request" response to our API client (the caller of DELETE /associatedDocuments/foo) means that that client sent us a bad request, not that we sent AWS a bad request, so given an SdkClientException, we should also send 500 to our API client.

lbodor: Generally, but probably won't end up mattering in this case, if you want to do different things for different exceptions, then you should register different exception handlers, rather than just one. This will eliminate calling "instance of" in the handler.

aws/deploy-local.sh
frankfu03: I'll do it in a separate PR as other scripts need to be changes as well.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
lbodor: I would register a single exception handler for all exceptions (Exception.class) and return 500.

lbodor: I wonder if that is already the default behavior.

frankfu03: Good, so now a single handler for all exceptions is registered which returns 500 only.

frankfu03:
@lbodor: Please let me know if I missed any comments. This PR seems work well except I am not sure if AWS profile gnss-metadata will work properly in DEV. Any more comments or approval?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -0,0 +1,121 @@
...
+    public ResponseEntity<String> upload(@RequestPart("file") MultipartFile file)
+            throws SdkClientException, AmazonServiceException, IOException {
+        String documentName = file.getOriginalFilename();
+        ObjectMetadata objectMetadata = new ObjectMetadata();
+        objectMetadata.setContentLength(file.getSize());
+        objectMetadata.setContentType(file.getContentType());
+
+        PutObjectRequest putObjectRequest = new PutObjectRequest(this.bucketName,
+                documentName, file.getInputStream(), objectMetadata)
+                .withCannedAcl(CannedAccessControlList.PublicRead);
```
lbodor, line 70: 4 spaces, not 8, leave blank lines around multi-line statements

PR #658, opened 2020-02-24
Remove unique name and fileReference constraints from sitelog_associateddocument table

frankfu03:


frankfu03:
As suggested, I added `deferrable="false", initiallyDeferred="true"` for the two unique field constraints in liquibase change log and it works fine. So, I'll created a new PR for deferring checking constraints until transactions are committed.

PR #662, opened 2020-03-20
Api for skeleton file request WIP

dotzhou:
1. Need to change RinexFileHeader to add one "observer agency" line
2. For all review and comments for assembler really working?

### DIFF
--- a/gws-core/pom.xml
+++ b/gws-core/pom.xml
@@ -125,11 +125,6 @@
             <version>0.1.0</version>
         </dependency>
         <dependency>
-            <groupId>au.gov.ga.gnss</groupId>
-            <artifactId>gnss-support-rinex</artifactId>
-            <version>1.0.0-SNAPSHOT</version>
-        </dependency>
-        <dependency>
             <groupId>au.gov.ga</groupId>
             <artifactId>xml-schemer</artifactId>
             <version>0.1.0</version>
--- a/gws-core/pom.xml
+++ b/gws-core/pom.xml
@@ -125,6 +125,11 @@
             <version>0.1.0</version>
         </dependency>
         <dependency>
+            <groupId>au.gov.ga.gnss</groupId>
+            <artifactId>gnss-support-rinex</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
             <groupId>au.gov.ga</groupId>
             <artifactId>xml-schemer</artifactId>
             <version>0.1.0</version>
--- a/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
+++ b/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
@@ -24,8 +24,7 @@ import java.util.Comparator;
 import java.util.List;
 
 @RestController
-@EnableEntityLinks
-@RequestMapping("/skeleton")
+@RequestMapping("/skeletonFiles")
 public class SkeletonEndpoint {
 
     private static final Logger log = LoggerFactory.getLogger(SkeletonEndpoint.class);
@@ -50,21 +49,16 @@ public class SkeletonEndpoint {
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         }
 
-        String markerName = fourCharId;
         String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
-
-        String agency = "Agency";
-        String observer = "Observer";
+        String agency = "";
         List<SiteResponsibleParty> siteContacts = siteLog.getSiteContacts();
         if (!siteContacts.isEmpty()) {
             agency = siteContacts.get(0).getParty().getOrganisationName().toString();
-            observer = siteContacts.get(0).getParty().getIndividualName();
         }
 
         RinexFileHeader rinexFileHeader = new RinexFileHeader();
-        rinexFileHeader.setMarkerName(markerName);
+        rinexFileHeader.setMarkerName(fourCharId);
         rinexFileHeader.setMarkerNumber(markerNumber);
-        rinexFileHeader.setObserver(observer);
         rinexFileHeader.setAgency(agency);
 
         List<GnssReceiverLogItem> receiverLogItemList = new ArrayList<>(siteLog.getGnssReceivers());
--- a/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
+++ b/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
@@ -1,52 +1,51 @@
 package au.gov.ga.geodesy.port.adapter.rest;
 
+import au.gov.ga.geodesy.domain.model.SiteResponsibleParty;
+import au.gov.ga.geodesy.domain.model.sitelog.GnssAntennaLogItem;
+import au.gov.ga.geodesy.domain.model.sitelog.GnssReceiverLogItem;
+import au.gov.ga.geodesy.domain.model.sitelog.SiteLog;
+import au.gov.ga.geodesy.domain.model.sitelog.SiteLogRepository;
+import au.gov.ga.gnss.support.rinex.RinexFileHeader;
+import com.vividsolutions.jts.geom.Point;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.hateoas.config.EnableEntityLinks;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody;
+
+import java.io.OutputStreamWriter;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import au.gov.ga.geodesy.domain.model.sitelog.GnssAntennaLogItem;
-import au.gov.ga.geodesy.domain.model.sitelog.GnssReceiverLogItem;
-import com.vividsolutions.jts.geom.Point;
-import org.apache.commons.lang3.tuple.MutableTriple;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.data.rest.webmvc.PersistentEntityResource;
-import org.springframework.data.rest.webmvc.PersistentEntityResourceAssembler;
-import org.springframework.data.rest.webmvc.RepositoryRestController;
-import org.springframework.hateoas.config.EnableEntityLinks;
-import org.springframework.http.HttpStatus;
-import org.springframework.http.ResponseEntity;
-import org.springframework.transaction.annotation.Transactional;
-import org.springframework.web.bind.annotation.*;
-
-import au.gov.ga.geodesy.domain.model.sitelog.SiteLog;
-import au.gov.ga.geodesy.domain.model.sitelog.SiteLogRepository;
-import au.gov.ga.gnss.support.rinex.RinexFileHeader;
-
-@RepositoryRestController
+@RestController
 @EnableEntityLinks
 @RequestMapping("/skeleton")
 public class SkeletonEndpoint {
 
     private static final Logger log = LoggerFactory.getLogger(SiteLogEndpoint.class);
+
     private static final String regex = "^(?<site>[A-Z0-9_]{4})([0-9]{2}[A-Z]{3})?\\.SKL$";
+
     private static final Pattern pattern = Pattern.compile(regex);
 
     @Autowired
     private SiteLogRepository siteLogs;
 
-    @Transactional
+    @Transactional(readOnly = true)
     @RequestMapping(
         value = "/{filename}",
         method = RequestMethod.GET,
         produces = "text/plain")
     @ResponseBody
-    public ResponseEntity<PersistentEntityResource> findByFourCharacterId(
-            @PathVariable String filename,
-            PersistentEntityResourceAssembler assembler) {
+    public ResponseEntity<StreamingResponseBody> findByFourCharacterId(
+            @PathVariable String filename) {
 
         String fourCharId = extractFourCharacterId(filename);
         if (fourCharId == null) {
@@ -54,69 +53,79 @@ public class SkeletonEndpoint {
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         }
 
-        try {
-            SiteLog siteLog = siteLogs.findByFourCharacterId(fourCharId);
-            if (siteLog == null) {
-                return new ResponseEntity<>(HttpStatus.NOT_FOUND);
-            }
-
-            /***
-             * SOLO                                                        MARKER NAME
-             * 51202M001                                                   MARKER NUMBER
-             *                     Geoscience Australia                    OBSERVER / AGENCY
-             * 5041K71018          TRIMBLE NETR9       5.44                REC # / TYPE / VERS
-             * 954                 JAVRINGANT_DM   NONE                    ANT # / TYPE
-             *  -5911340.1240  2156887.2990 -1038664.0050                  APPROX POSITION XYZ
-             *         0.0010        0.0000        0.0000                  ANTENNA: DELTA H/E/N
-             *                                                             END OF HEADER
-             */
-
-            String markerName = fourCharId;
-            String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
-            String observerAgency = "Geoscience Australia";
-
-            RinexFileHeader rinexFileHeader = new RinexFileHeader();
-            rinexFileHeader.setMarkerName(markerName);
-            rinexFileHeader.setMarkerNumber(markerNumber);
-            // rinexFileHeader.setObserverAgency(observerAgency);
-
-            List<GnssReceiverLogItem> receiverLogItemList =
-                    new ArrayList<GnssReceiverLogItem>(siteLog.getGnssReceivers());
-            receiverLogItemList.sort(Comparator.comparing(GnssReceiverLogItem::getDateInstalled).reversed());
-
-            GnssReceiverLogItem receiverLogItem = receiverLogItemList.get(0);
-            rinexFileHeader.setReceiverSerialNumber(receiverLogItem.getSerialNumber());
-            rinexFileHeader.setReceiverType(receiverLogItem.getType());
-            rinexFileHeader.setReceiverFirmwareVersion(receiverLogItem.getFirmwareVersion());
-
-            List<GnssAntennaLogItem> antennaLogItemList = new ArrayList<GnssAntennaLogItem>(siteLog.getGnssAntennas());
-            antennaLogItemList.sort(Comparator.comparing(GnssAntennaLogItem::getDateInstalled).reversed());
-
-            GnssAntennaLogItem antennaLogItem = antennaLogItemList.get(0);
-            rinexFileHeader.setAntennaSerialNumber(antennaLogItem.getSerialNumber());
-            rinexFileHeader.setAntennaType(antennaLogItem.getType());
-
-            Point position = siteLog.getSiteLocation().getApproximatePosition().getCartesianPosition();
-            rinexFileHeader.setApproxPositionXyz(new MutableTriple<Double, Double, Double>(
-                    position.getCoordinate().x, position.getCoordinate().y, position.getCoordinate().z));
-
-            rinexFileHeader.setAntennaMarkerArp(new MutableTriple<Double, Double, Double>(
-                    antennaLogItem.getMarkerArpUpEcc(),
-                    antennaLogItem.getMarkerArpEastEcc(),
-                    antennaLogItem.getMarkerArpNorthEcc()));
-
-            return ResponseEntity.ok(assembler.toResource(rinexFileHeader));
-        } catch (Exception e) {
-            // In case of invalid site log file, no receiver or no antenna records at all, etc.
-            log.error("Invalid siteLogItem for site " + fourCharId);
+        SiteLog siteLog = siteLogs.findByFourCharacterId(fourCharId);
+        if (siteLog == null) {
+            log.error("Failed to retrive site log for station " + fourCharId);
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         }
+
+        /***
+         * SOLO                                                        MARKER NAME
+         * 51202M001                                                   MARKER NUMBER
+         *                     Geoscience Australia                    OBSERVER / AGENCY
+         * 5041K71018          TRIMBLE NETR9       5.44                REC # / TYPE / VERS
+         * 954                 JAVRINGANT_DM   NONE                    ANT # / TYPE
+         *  -5911340.1240  2156887.2990 -1038664.0050                  APPROX POSITION XYZ
+         *         0.0010        0.0000        0.0000                  ANTENNA: DELTA H/E/N
+         *                                                             END OF HEADER
+         */
+
+        String markerName = fourCharId;
+        String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
+
+        String agency = "N/A";
+        String observer = "N/A";
+        List<SiteResponsibleParty> siteContacts = siteLog.getSiteContacts();
+        if (!siteContacts.isEmpty()) {
+            agency = siteContacts.get(0).getParty().getOrganisationName().toString();
+            observer = siteContacts.get(0).getParty().getIndividualName();
+        }
+
+        RinexFileHeader rinexFileHeader = new RinexFileHeader();
+        rinexFileHeader.setMarkerName(markerName);
+        rinexFileHeader.setMarkerNumber(markerNumber);
+        rinexFileHeader.setObserver(observer);
+        rinexFileHeader.setAgency(agency);
+
+        List<GnssReceiverLogItem> receiverLogItemList = new ArrayList<>(siteLog.getGnssReceivers());
+        if (receiverLogItemList.isEmpty()) {
+            log.error("No receiver record for station " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+        receiverLogItemList.sort(Comparator.comparing(GnssReceiverLogItem::getDateInstalled).reversed());
+
+        GnssReceiverLogItem receiverLogItem = receiverLogItemList.get(0);
+        rinexFileHeader.setReceiverSerialNumber(receiverLogItem.getSerialNumber());
+        rinexFileHeader.setReceiverType(receiverLogItem.getType());
+        rinexFileHeader.setReceiverFirmwareVersion(receiverLogItem.getFirmwareVersion());
+
+        List<GnssAntennaLogItem> antennaLogItemList = new ArrayList<>(siteLog.getGnssAntennas());
+        if (antennaLogItemList.isEmpty()) {
+            log.error("No antenna record for station " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+        antennaLogItemList.sort(Comparator.comparing(GnssAntennaLogItem::getDateInstalled).reversed());
+
+        GnssAntennaLogItem antennaLogItem = antennaLogItemList.get(0);
+        rinexFileHeader.setAntennaSerialNumber(antennaLogItem.getSerialNumber());
+        rinexFileHeader.setAntennaType(antennaLogItem.getType());
+
+        Point position = siteLog.getSiteLocation().getApproximatePosition().getCartesianPosition();
+        rinexFileHeader.getApproxPositionXyz().setLeft(position.getCoordinate().x);
+        rinexFileHeader.getApproxPositionXyz().setMiddle(position.getCoordinate().y);
+        rinexFileHeader.getApproxPositionXyz().setRight(position.getCoordinate().z);
+
+        rinexFileHeader.getAntennaMarkerArp().setLeft(antennaLogItem.getMarkerArpUpEcc());
+        rinexFileHeader.getAntennaMarkerArp().setMiddle(antennaLogItem.getMarkerArpEastEcc());
+        rinexFileHeader.getAntennaMarkerArp().setRight(antennaLogItem.getMarkerArpNorthEcc());
+
+        return ResponseEntity.ok(outputStream -> rinexFileHeader.write(new OutputStreamWriter(outputStream)));
     }
 
     /**
      * Extract the four character id from skeleton file request.
      */
-    String extractFourCharacterId(String identifier) {
+    private String extractFourCharacterId(String identifier) {
         Matcher matcher = pattern.matcher(identifier);
         if (matcher.find()) {
             return matcher.group("site");
--- a/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
+++ b/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
@@ -11,7 +11,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.hateoas.config.EnableEntityLinks;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.bind.annotation.*;
@@ -27,21 +29,21 @@ import java.util.List;
 @RequestMapping("/skeleton")
 public class SkeletonEndpoint {
 
-    private static final Logger log = LoggerFactory.getLogger(SiteLogEndpoint.class);
+    private static final Logger log = LoggerFactory.getLogger(SkeletonEndpoint.class);
 
     @Autowired
     private SiteLogRepository siteLogs;
 
     @Transactional(readOnly = true)
     @RequestMapping(
-        value = "/{filename:[a-zA-Z0-9_]{4}([0-9]{2}[a-zA-Z]{3})?\\.((SKL)|(skl))}",
+        value = "/{filename:[a-zA-Z0-9_]{4}(?:[0-9]{2}[a-zA-Z]{3})?\\.(?:(?:SKL)|(?:skl))}",
         method = RequestMethod.GET,
         produces = "text/plain")
     @ResponseBody
     public ResponseEntity<StreamingResponseBody> findByFourCharacterId(
-            @PathVariable String filename) {
+        @PathVariable String filename) {
 
-        String fourCharId = filename.substring(0, 4);
+        String fourCharId = filename.substring(0, 4).toUpperCase();
 
         SiteLog siteLog = siteLogs.findByFourCharacterId(fourCharId);
         if (siteLog == null) {
@@ -98,6 +100,11 @@ public class SkeletonEndpoint {
         rinexFileHeader.getAntennaMarkerArp().setMiddle(antennaLogItem.getMarkerArpEastEcc());
         rinexFileHeader.getAntennaMarkerArp().setRight(antennaLogItem.getMarkerArpNorthEcc());
 
-        return ResponseEntity.ok(outputStream -> rinexFileHeader.write(new OutputStreamWriter(outputStream)));
+        HttpHeaders httpHeaders = new HttpHeaders();
+        httpHeaders.setContentType(MediaType.TEXT_PLAIN);
+
+        return ResponseEntity.ok()
+            .headers(httpHeaders)
+            .body(outputStream -> rinexFileHeader.write(new OutputStreamWriter(outputStream)));
     }
 }
--- a/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
+++ b/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
@@ -40,7 +40,7 @@ public class SkeletonEndpoint {
 
     @Transactional(readOnly = true)
     @RequestMapping(
-        value = "/{filename}",
+        value = "/{filename:[a-zA-Z0-9_]{4}([0-9]{2}[a-zA-Z]{3})?\\.((SKL)|(skl))}",
         method = RequestMethod.GET,
         produces = "text/plain")
     @ResponseBody
@@ -50,7 +50,7 @@ public class SkeletonEndpoint {
         String fourCharId = extractFourCharacterId(filename);
         if (fourCharId == null) {
             log.error("Invalid skeleton file request received: " + filename);
-            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
         }
 
         SiteLog siteLog = siteLogs.findByFourCharacterId(fourCharId);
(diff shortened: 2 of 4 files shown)


gws-core/pom.xml
```diff
@@ -124,6 +124,11 @@
             <artifactId>igssitelog-bindings</artifactId>
             <version>0.1.0</version>
         </dependency>
+        <dependency>
+            <groupId>au.gov.ga.gnss</groupId>
+            <artifactId>gnss-support-rinex</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
+        </dependency>
```
lbodor: For this to work, we will need a `<repositories>` tag to add maven-artifactory-prod to the list of repositories that maven will search through to find dependencies. See geodesy-archive/maven-parent/pom.xml.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+public class SkeletonEndpoint {
+
+    private static final Logger log = LoggerFactory.getLogger(SiteLogEndpoint.class);
+    private static final String regex = "^(?<site>[A-Z0-9_]{4})([0-9]{2}[A-Z]{3})?\\.SKL$";
+    private static final Pattern pattern = Pattern.compile(regex);
+
+    @Autowired
+    private SiteLogRepository siteLogs;
+
+    @Transactional
```
lbodor: @Transactional(readOnly = true) will run quicker and ensure that we don't unintentionally save anything to the database.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+import au.gov.ga.gnss.support.rinex.RinexFileHeader;
+
+@RepositoryRestController
+@EnableEntityLinks
+@RequestMapping("/skeleton")
+public class SkeletonEndpoint {
+
+    private static final Logger log = LoggerFactory.getLogger(SiteLogEndpoint.class);
+    private static final String regex = "^(?<site>[A-Z0-9_]{4})([0-9]{2}[A-Z]{3})?\\.SKL$";
+    private static final Pattern pattern = Pattern.compile(regex);
```
umeat: Is it more common to have newlines between class attributes? Either way is probably fine but I prefer the additional white space, personally.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Autowired
+    private SiteLogRepository siteLogs;
+
+    @Transactional
+    @RequestMapping(
+        value = "/{filename}",
```
lbodor: You could probably use a regular expression in the @RequestMapping `value` attribute, so that you don't have to handle invalid values yourself.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: I will do your way! Thanks

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+             */
+
+            String markerName = fourCharId;
+            String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
+            String observerAgency = "Geoscience Australia";
+
+            RinexFileHeader rinexFileHeader = new RinexFileHeader();
+            rinexFileHeader.setMarkerName(markerName);
+            rinexFileHeader.setMarkerNumber(markerNumber);
+            // rinexFileHeader.setObserverAgency(observerAgency);
```
umeat: Commented code.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Learnt now. I will change it now. Thanks

gws-core/pom.xml
dotzhou: Not sure how to do it yet? Need your advises

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+
+            rinexFileHeader.setAntennaMarkerArp(new MutableTriple<Double, Double, Double>(
+                    antennaLogItem.getMarkerArpUpEcc(),
+                    antennaLogItem.getMarkerArpEastEcc(),
+                    antennaLogItem.getMarkerArpNorthEcc()));
+
+            return ResponseEntity.ok(assembler.toResource(rinexFileHeader));
+        } catch (Exception e) {
+            // In case of invalid site log file, no receiver or no antenna records at all, etc.
+            log.error("Invalid siteLogItem for site " + fourCharId);
```
umeat: Doesn't look like the exception gets logged. You should log the exception so we know what was invalid.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: The field name is "Observer / Agency", so this property should probably be `observerOrAgency` or `observer` or `agency`. Is that right?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Good idea. Try to implement it

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+            String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
+            String observerAgency = "Geoscience Australia";
+
+            RinexFileHeader rinexFileHeader = new RinexFileHeader();
+            rinexFileHeader.setMarkerName(markerName);
+            rinexFileHeader.setMarkerNumber(markerNumber);
+            // rinexFileHeader.setObserverAgency(observerAgency);
+
+            List<GnssReceiverLogItem> receiverLogItemList =
+                    new ArrayList<GnssReceiverLogItem>(siteLog.getGnssReceivers());
```
lbodor: Use the diamon operator `<>`.
```
List<GnssReceiverLogItem> receiverLogItemList = new ArrayList<>(siteLog.getGnssReceivers());
```

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: I believe your way "observerOrAgency" is better.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+            receiverLogItemList.sort(Comparator.comparing(GnssReceiverLogItem::getDateInstalled).reversed());
```
lbodor: You should try using Setup here so that you don't have to find the current equipment.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: We can't catch all exceptions and then say "Invalid siteLogItem...", because the exception could be anything. I think we should log the exception without attempting to interpret what went wrong and sent HTTP 500 to the client.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+            // In case of invalid site log file, no receiver or no antenna records at all, etc.
+            log.error("Invalid siteLogItem for site " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+    }
+
+    /**
+     * Extract the four character id from skeleton file request.
+     */
+    String extractFourCharacterId(String identifier) {
```
lbodor: private

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.bind.annotation.*;
+
+import au.gov.ga.geodesy.domain.model.sitelog.SiteLog;
+import au.gov.ga.geodesy.domain.model.sitelog.SiteLogRepository;
+import au.gov.ga.gnss.support.rinex.RinexFileHeader;
+
+@RepositoryRestController
```
lbodor: Probably just `@RestController` would be more appropriate, because there isn't a repository for skeleton files.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,126 @@
...
+            Point position = siteLog.getSiteLocation().getApproximatePosition().getCartesianPosition();
+            rinexFileHeader.setApproxPositionXyz(new MutableTriple<Double, Double, Double>(
+                    position.getCoordinate().x, position.getCoordinate().y, position.getCoordinate().z));
+
+            rinexFileHeader.setAntennaMarkerArp(new MutableTriple<Double, Double, Double>(
+                    antennaLogItem.getMarkerArpUpEcc(),
+                    antennaLogItem.getMarkerArpEastEcc(),
+                    antennaLogItem.getMarkerArpNorthEcc()));
+
+            return ResponseEntity.ok(assembler.toResource(rinexFileHeader));
```
lbodor: Entity assemblers are for Spring Data/JPA entities. Here we should use `rinexFileHeader.write` method to generate header text. One option would be
1) Set this method's return type to `StreamingResponseBody`
2) Remove `PersistentEntityResourceAssembler` argument
3) Return `outputStream -> rinexFileHeader.write(new WriterStream(outputStream))`

If you include a regular expression in @RequestMapping and if you register an exception handler to return HTTP 500, then you don't need to handle anything other than the happy-day scenario in the body of the method. See other endpoint classes for an example of an exception handler.

lbodor: The arrow in 3) above is special syntax for returning a lambda, which is type-compatable with any class in java that is annotated with `@FunctionalInterface`. StreamingResponseBody is a functional interface. For a more verbose example, see: https://www.logicbig.com/tutorials/spring-framework/spring-web-mvc/streaming-response-body.html.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Sorry, forget it.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: I tried to use Setup. But I failed to retrieved the information I need. I try to ask you how to retrieve those information later

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Yes, it should be private

gws-core/pom.xml
lbodor: gnss-support-rinex jar file is already in s3 at https://gnss-maven-artifactory-prod.s3-ap-southeast-2.amazonaws.com/releases/au/gov/ga/gnss/gnss-support-rinex/1.0.0/gnss-support-rinex-1.0.0.jar

You need to tell maven about this repository using the <repositories>...</repositories> tag in pom.xml so that it can be downloaded and used.

See geodesy-archive/maven-parent/pom.xml for an example

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Yes, RestController is more appropriate

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Really thanks for that link

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: I think we should try use setups in this PR. We can fetch both the sitelog for some stuff and the setup for the equipment.

gws-core/pom.xml
dotzhou: one example at 

gws-parent/pom.xml

 53     <repositories>
 54         <repository>
 55             <id>gnss-artifactory-release</id>
 56             <name>S3 GNSS Artifactory</name>
 57             <url>s3://gnss-artifactory/maven-releases/</url>
 58         </repository>
 59         <repository>
 60             <id>gnss-artifactory-snapshot</id>
 61             <name>S3 GNSS Artifactory</name>
 62             <url>s3://gnss-artifactory/maven-snapshots/</url>
 63         </repository>
 64     </repositories>

dotzhou: ted@soft:~/.m2/repository/au/gov/ga/gnss/gnss-support-rinex/1.0.0-SNAPSHOT$ ls
gnss-support-rinex-1.0.0-SNAPSHOT.jar  gnss-support-rinex-1.0.0-SNAPSHOT.pom  maven-metadata-local.xml  _remote.repositories  resolver-status.properties
ted@soft:~/.m2/repository/au/gov/ga/gnss/gnss-support-rinex/1.0.0-SNAPSHOT$ pwd


Not sure through ${HOME}/.m2/respository is enough or not ?

dotzhou: ted@soft:~/.m2/repository/au/gov/ga/gnss/gnss-support-rinex/1.0.0-SNAPSHOT$ ls 
total 44
-rw-r--r-- 1 ted users  3144 Mar 17 16:29 gnss-support-rinex-1.0.0-SNAPSHOT.pom
-rw-r--r-- 1 ted users 16855 Mar 18 12:13 gnss-support-rinex-1.0.0-SNAPSHOT.jar
-rw-r--r-- 1 ted users   221 Mar 18 12:15 _remote.repositories
-rw-r--r-- 1 ted users   718 Mar 18 12:15 maven-metadata-local.xml
drwxr-xr-x 3 ted users  4096 Mar 18 12:15 ..
drwxr-xr-x 2 ted users  4096 Mar 19 09:18 .
-rw-r--r-- 1 ted users  1838 Mar 19 09:18 resolver-status.properties
ted@soft:~/.m2/repository/au/gov/ga/gnss/gnss-support-rinex/1.0.0-SNAPSHOT$

dotzhou: Try to change

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Will be replaced.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Done now, please review

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Implemented as you suggested. Please review

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Done, please review

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: This comment is still relevant.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+        method = RequestMethod.GET,
+        produces = "text/plain")
+    @ResponseBody
+    public ResponseEntity<StreamingResponseBody> findByFourCharacterId(
+            @PathVariable String filename) {
+
+        String fourCharId = extractFourCharacterId(filename);
+        if (fourCharId == null) {
+            log.error("Invalid skeleton file request received: " + filename);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
lbodor: 400 for invalid request

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+         *  -5911340.1240  2156887.2990 -1038664.0050                  APPROX POSITION XYZ
+         *         0.0010        0.0000        0.0000                  ANTENNA: DELTA H/E/N
+         *                                                             END OF HEADER
+         */
+
+        String markerName = fourCharId;
+        String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
+
+        String agency = "N/A";
+        String observer = "N/A";
```
lbodor: Why "N/A"?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+        RinexFileHeader rinexFileHeader = new RinexFileHeader();
+        rinexFileHeader.setMarkerName(markerName);
+        rinexFileHeader.setMarkerNumber(markerNumber);
+        rinexFileHeader.setObserver(observer);
+        rinexFileHeader.setAgency(agency);
+
+        List<GnssReceiverLogItem> receiverLogItemList = new ArrayList<>(siteLog.getGnssReceivers());
+        if (receiverLogItemList.isEmpty()) {
+            log.error("No receiver record for station " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
lbodor: Should this be 500?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+
+        GnssReceiverLogItem receiverLogItem = receiverLogItemList.get(0);
+        rinexFileHeader.setReceiverSerialNumber(receiverLogItem.getSerialNumber());
+        rinexFileHeader.setReceiverType(receiverLogItem.getType());
+        rinexFileHeader.setReceiverFirmwareVersion(receiverLogItem.getFirmwareVersion());
+
+        List<GnssAntennaLogItem> antennaLogItemList = new ArrayList<>(siteLog.getGnssAntennas());
+        if (antennaLogItemList.isEmpty()) {
+            log.error("No antenna record for station " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
lbodor: Should this be 500?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: This comment is still relevant.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+
+        String markerName = fourCharId;
+        String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
+
+        String agency = "N/A";
+        String observer = "N/A";
+        List<SiteResponsibleParty> siteContacts = siteLog.getSiteContacts();
+        if (!siteContacts.isEmpty()) {
+            agency = siteContacts.get(0).getParty().getOrganisationName().toString();
+            observer = siteContacts.get(0).getParty().getIndividualName();
```
lbodor: @umeat, what do you think about using site contact individual name as the observer?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+        /***
+         * SOLO                                                        MARKER NAME
+         * 51202M001                                                   MARKER NUMBER
+         *                     Geoscience Australia                    OBSERVER / AGENCY
+         * 5041K71018          TRIMBLE NETR9       5.44                REC # / TYPE / VERS
+         * 954                 JAVRINGANT_DM   NONE                    ANT # / TYPE
+         *  -5911340.1240  2156887.2990 -1038664.0050                  APPROX POSITION XYZ
+         *         0.0010        0.0000        0.0000                  ANTENNA: DELTA H/E/N
+         *                                                             END OF HEADER
+         */
```
lbodor: Do we need this comment here?

dotzhou: Remove those comments

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Open for your feedback

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Try to do

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Try to do

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: @umean, do we want to be generating rinex files without an observer and agency?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Default "observer" and default "agency", open for all of your feedback.

dotzhou: N/A --> Not Available.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Do your way

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: From Site Log we could get all information we need.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: Blank is better than "n/a" ar "not available", in my opinion. Could we use "Goescience Australia" as the default? Do we only generate rinex files for GAs sites?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,135 @@
...
+import java.io.OutputStreamWriter;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@RestController
+@EnableEntityLinks
+@RequestMapping("/skeleton")
```
lbodor: Is "skeleton" descriptive enough as a top-level endpoint in GWS?
```
/siteLogs
/corsSites
/gnssReceivers
/skeleton
```

lbodor: `/rinexHeaders/alic`?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Should be there

lbodor: I don't understand, what should be where?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Some are not GA's sites.

Should I let the agency default to "Geoscience Australia", and default observer to empty ""?

If both observer and agency are empty, not sure the consequence?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: If both "observer" and "agency" are empty, not sure the consequence?

Brandon you have to decide the default text for both "observer" or "agency"?

How about default "observer" = "observer", and "agency" = "agency"?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final String regex = "^(?<site>[A-Z0-9_]{4})([0-9]{2}[A-Z]{3})?\\.SKL$";
+
+    private static final Pattern pattern = Pattern.compile(regex);
+
+    @Autowired
+    private SiteLogRepository siteLogs;
+
+    @Transactional(readOnly = true)
+    @RequestMapping(
+        value = "/{filename:[a-zA-Z0-9_]{4}([0-9]{2}[a-zA-Z]{3})?\\.((SKL)|(skl))}",
```
lbodor: great

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,124 @@
...
+        method = RequestMethod.GET,
+        produces = "text/plain")
+    @ResponseBody
+    public ResponseEntity<StreamingResponseBody> findByFourCharacterId(
+            @PathVariable String filename) {
+
+        String fourCharId = extractFourCharacterId(filename);
```
lbodor: I don't think you need `extractFourCharacterId` any more. The point of adding the regex in the mapping was to avoid having to check the filename yourself and handle the error condition. Is that not working?

lbodor: I think you should add a test to this PR to test your code. Take a look at some integration tests in the code base that use spring mock-mvc and rest-assured to the test various endpoints.

dotzhou: Thanks to Lazar.

I try to clean it.

I also need to add integration test now. Thanks to your suggestions for how to test.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,103 @@
...
+    @Autowired
+    private SiteLogRepository siteLogs;
+
+    @Transactional(readOnly = true)
+    @RequestMapping(
+        value = "/{filename:[a-zA-Z0-9_]{4}([0-9]{2}[a-zA-Z]{3})?\\.((SKL)|(skl))}",
+        method = RequestMethod.GET,
+        produces = "text/plain")
+    @ResponseBody
+    public ResponseEntity<StreamingResponseBody> findByFourCharacterId(
```
lbodor, line 41: I see now that this should be site id, not four character id.

lbodor: Actually, it's not clear to me yet what it should be. You are accepting 9 character ids, but you are searching by the first four characters.

dotzhou: Designed four end points as below:

/skeleton/alic.skl
/skeleton/ALIC.SKL
/skeleton/alic00aus.skl
/skeleton/ALIC00AUS.skl

They should  respond with same skeleton header.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: I don't think that "Agency" and "Observer" are valid default values. I think we should either go with blanks, or something that means something like "Geoscience Australia". @umeat?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: Shouldn't the resource endpoint be plural?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,109 @@
...
+import org.springframework.web.bind.annotation.*;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+@RestController
+@EnableEntityLinks
```
lbodor, line 27: Is @EnableEntityLinks needed? I suspect that it isn't.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: This is still relevant.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
lbodor: This is still relevant.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Yes, I believe it is not relevant. I remove

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpointITest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        assertThat(text, containsString("AGENCY"));
+        assertThat(text, containsString("50137M001"));
+    }
+
+    @Test
+    public void testNotFound() throws Exception {
+        given()
+            .when()
+            .get("/skeleton/0000.SKL")
```
lbodor, line 65: What happens when the filename does not match the regular expression given in the @RequestMapping annotation?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,109 @@
...
+
+        String fourCharId = filename.substring(0, 4).toUpperCase();
+
+        SiteLog siteLog = siteLogs.findByFourCharacterId(fourCharId);
+        if (siteLog == null) {
+            log.error("Failed to retrive site log for station " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+
+        String markerName = fourCharId;
```
lbodor, line 53: The above line is unnecessary, we should just say `rinexFileHeader.setMarkerName(fourCharId)` further down.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpointITest.java
dotzhou: I really have to check the documentation now?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Good idea

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpointITest.java
lbodor: You should add another test method to find out.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: then the valid request paths are as below:

/skeletonFiles/ALIC.SKL
/skeletonFiles/alic.skl
/skeletonFiles/ALIC00AUS.SKL
/skeletonFiles/alic00aus.skl

BNC did this way:
http://www.epncb.oma.be:80/stations/log/skl/brus.skl 

/skl   is the path, skl seems to be a good choice?

dotzhou: Probably It is not rinexFileHeader since no Rinex version defined in the header

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpointITest.java
dotzhou: Good idea. Try to do

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
umeat: @dotzhou can we make this `skeletonFiles`?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
umeat: Looks like this is still relevant

dotzhou: if you don't mind. I keep as it is for better readability? Easy to check against Skeleton Header.

lbodor: Conceptually, I find it the same
```
String markerName = fourCharId;
rinexHeaderFile.setMarkerName(markerName);
```
and
```
rinexFileHeader.setMarkerName(fourCharId)
```

If `String markerName = fourCharId` is OK, then why isn't `rinexFileHeader.setMarkerName(fourCharId)`?

dotzhou: Really sorry for this. Not my intention

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: /skeletonFiles/ is the path

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Hi, Brandon:

1. If you don't want Ryan's name in the header, what should we put in the field of "Observer"? Default to be empty "".

All "Obersver" left as empty, if you are happy with current state.


The "Agency" will be filled up as "Geoscience Australia" for our own station.
"Agency" default to be empty "".

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
umeat: @dotzhou have you committed that change? Because I still see "/skeleton".

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
umeat: I think we should default to empty strings. The other two options are to fail, or to return a header without that line. I'm not sure what BNC will do if the line is not present in the skeleton file. Probably worth testing, but I'm fine with Observer being blank since it's blank in all of our RINEX files submitted from site.

dotzhou:
Brandon, Lazar: 

Need your final review now. I could start to do another task for Rinex 3 file generation on Brandon's branch.

Thanks for your time

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -0,0 +1,103 @@
...
+            log.error("Failed to retrive site log for station " + fourCharId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+
+        String markerNumber = siteLog.getSiteIdentification().getIersDOMESNumber();
+        String agency = "";
+        List<SiteResponsibleParty> siteContacts = siteLog.getSiteContacts();
+        if (!siteContacts.isEmpty()) {
+            agency = siteContacts.get(0).getParty().getOrganisationName().toString();
+        }
```
lbodor, line 57: Always use `null` instead of the empty string for an absent value. But, event better, move the if-statement further down to avoid the local `agency` variable altogether, just like you're already doing for receivers and antennas. Same applies for `markerNumber`.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpointITest.java
```diff
@@ -0,0 +1,78 @@
...
+            .get("/skeletonFiles/0000.SKL")
+            .then()
+            .statusCode(HttpStatus.NOT_FOUND.value());
+    }
+
+    @Test
+    public void testInvalid() throws Exception {
+        given()
+            .when()
+            .get("/skeletonFiles/123456789.xyz")
```
lbodor, line 74: Great, the regex works.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
dotzhou: Try to change. 

If both observer and agency are null, not sure will break RinexHeader class?

PR #664, opened 2020-04-06
Nine character id for Mark name

dotzhou:
I recommends nine character id for "MARKER NAME"

### DIFF
--- a/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
+++ b/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
@@ -50,7 +50,12 @@ public class SkeletonEndpoint {
         }
 
         RinexFileHeader rinexFileHeader = new RinexFileHeader();
-        rinexFileHeader.setMarkerName(fourCharId);
+
+        if (filename.length() == 13) {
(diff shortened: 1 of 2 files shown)


gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -50,7 +50,12 @@
         }
 
         RinexFileHeader rinexFileHeader = new RinexFileHeader();
-        rinexFileHeader.setMarkerName(fourCharId);
+
+        if (filename.length() == 13) {
+            rinexFileHeader.setMarkerName(filename.substring(0, 9).toUpperCase());
+        } else {
+            rinexFileHeader.setMarkerName(fourCharId);
```
umeat, line 57: Does it make sense for this to be conditional? Should we try to derive the 9 character ID in all cases?

dotzhou: It is a better idea.

I can retrieve country code from GNSS Site Manager api.

However, it is not easy get receiver number? 0 or 1

umeat:
We should definitely speak to the analysis team (and as many people as possible really) about this before doing anything.

PR #665, opened 2020-04-07
Change to return document url embedded in response body

frankfu03:
Have to roll back part of the changes made in previous PR because that we could not retrieve the url of uploaded documents from response header location in GNSS-SITE-MANAGER. I'll attach snapshots showing the difference in Slack for reference.


frankfu03:
I have added "location" to access-control-expose-headers and it seems work fine. I'll submit a separate PR and close this one.

PR #666, opened 2020-04-07
Add "location" to access control expose header list

frankfu03:


gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -76,6 +78,9 @@ public AssociatedDocumentEndpoint() {
         log.info("Uploaded " + documentName + " to S3 bucket: " + objectUrl);
         HttpHeaders responseHeaders = new HttpHeaders();
         responseHeaders.setLocation(new URI(objectUrl));
+        List<String> headerNames = new ArrayList<>();
+        headerNames.add("location");
+        responseHeaders.setAccessControlExposeHeaders(headerNames);
```
lbodor, line 83: Would it also work to make this adjustment to exposed headers in https://github.com/GeoscienceAustralia/Geodesy-Web-Services/blob/master/gws-webapp/src/main/java/au/gov/ga/geodesy/support/spring/WebAppInitializer.java, so that it's all in one place?

frankfu03: Yes, it also works after moving the code to WebAppInitializer.java.

gws-webapp/src/main/java/au/gov/ga/geodesy/support/spring/WebAppInitializer.java
```diff
@@ -31,6 +31,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             rsp.setHeader("Access-Control-Allow-Origin", "*");
             rsp.setHeader("Access-Control-Allow-Methods", "POST, PUT, GET, OPTIONS, DELETE, PATCH");
             rsp.setHeader("Access-Control-Allow-Headers", "origin, content-type, accept, x-requested-with, authorization");
+            rsp.setHeader("Access-Control-Expose-Headers", "location");
```
lbodor, line 34: Does this overwrite the headers that are exposed by default? If so, I worry that something may break somewhere. You were getting 3 headers in that screenshot you pasted from the site-manager debug console. Why don't we just add the location to those three. In fact, I recall that six response headers were exposed by defaut.

lbodor: From the post I referenced last time: CORS requests only expose 6 safelisted headers : Cache-Control Content-Language Content-Type Expires Last-Modified & Pragma. (https://stackoverflow.com/questions/52443706/angular-httpclient-missing-response-headers). Could you try printing the the value of the exposed-headers header before you overwrite it? Or find some other way to confirm what the defaults should be.

lbodor: I think that we should leave the defaults in tact.

frankfu03: I'll double check it. But it seems that it does not overwrite the headers that are exposed by default. It adds extra headers to the expose header list, which includes the 6 default safelisted headers. The PR passed all integration tests. 

The [documentation ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) states: A list of exposed headers consisting of zero or more header names other than the CORS-safelisted request headers that the resource might use and can be exposed.

lbodor: Great. That's good information, thanks for looking it up.

frankfu03: Thanks for your support as well.

PR #667, opened 2020-04-08
Fix unset variables error in aws/deploy.sh

frankfu03:
Alternative solution is to remove "-u" from "set -eEuo pipefail". However, unset variables are a common cause of bugs in shell scripts, so having unset variables cause an immediate exit is often highly desirable behavior.

### DIFF
--- a/aws/deploy.sh
+++ b/aws/deploy.sh
@@ -14,6 +14,7 @@ function usage {
     exit 1
 }
 
+dryRun=""
 while [[ $# -gt 0 ]]; do
     case $1 in
         -d|--dry-run )


umeat:
What does the capital E do in `set -eEou ...`? We typically use `-eou`.

frankfu03:
"-E":  If set, the ERR trap is inherited by shell functions. Trap allows you to catch signals and execute code when they occur.

aws/deploy.sh
```diff
@@ -14,6 +14,7 @@ function usage {
     exit 1
 }
 
+dryRun=""
```
lbodor, line 17: Just to note that because everything in bash is a string, `dryRun=` is the same as `dryRun=""`

frankfu03: I tested: both work fine. But the first one looks like missing something.

lbodor: I know, it looks a bit strange.

PR #668, opened 2020-04-14
Update dependency versions for projects related to upgrade of geodesyml v0.5

frankfu03:


gws-core/pom.xml
```diff
@@ -127,12 +127,12 @@
         <dependency>
             <groupId>au.gov.ga.geodesy</groupId>
             <artifactId>igssitelog-bindings</artifactId>
-            <version>0.1.0</version>
+            <version>0.2.0-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>au.gov.ga</groupId>
             <artifactId>xml-schemer</artifactId>
-            <version>0.1.0</version>
```
lbodor, line 135: Do we need to bring in any changes in xml-schemer?

gws-core/pom.xml
```diff
@@ -127,12 +127,12 @@
         <dependency>
             <groupId>au.gov.ga.geodesy</groupId>
             <artifactId>igssitelog-bindings</artifactId>
-            <version>0.1.0</version>
```
lbodor, line 130: Do we need to bring in any changes here?

gws-core/pom.xml
frankfu03: Changes are not related to upgrade of geodesyml v.0.5. The change is mainly the deploy of source code to s3-gnss-artifactory.

gws-core/pom.xml
frankfu03: The geodesyml version in igssitelog-bindings has been upgraded from v.0.4 to v.0.5, so it may need to upgrade its repo version. But if it is not used any more, we may just simply remove this dependency.

lbodor: Is there really a version of geodesyml in igssitelog-bindings. I didn't think that igssitelog-bindings depended on geodesyml at all?

PR #672, opened 2020-04-22
Release 0.2.2

lbodor:
New release of GWS, which grabs the now released dependencies at version 0.2.0 (using GeodesyML 0.4) and 0.3.0 of igssitelog-java-bindings, which does not depend on GeodesyML.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ services:
 - docker
 branches:
   only:
-  - master
-  - next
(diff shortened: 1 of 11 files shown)


lbodor:
I've updated this PR so that we deploy updates to geodesyml-0.4 branch itself, rather than to deploy updates to branches partaking in PRs against geodesyml-0.4.

PR #673, opened 2020-04-22
Change the expected setups.type to be "CorsSetup" in SetupSystemTest

lbodor:


lbodor:
This PR cherry-picks on old commit against master that fixes a broken system test.

PR #676, opened 2020-04-27
Deploy Prod 0.2.2 to production

lbodor:
- Deploy maven artifacts to s3
- Bump some dependency versions
- Fix a broken system test
- Add "test" and "prod" deployment trigger branches
- Delete obsolete site log test files


lbodor:
This release is the final geodesyml-0.4 release, which is api-compatible with all downstream components: converters, site-manager, and gnss-data, and we aren't expecting any outage.

PR #690, opened 2020-05-11
Skeleton endpoint to use Metadata Custodian

umeat:
Also includes change merged into master from recent prod release branch.


umeat:
I agree. Does this involve much more than just setting the version in the root pom file to 0.4.0.SNAPSHOT?

Snapshot is meant to be latest unstable, right?

lbodor:
Yes, 0.4.0-SNAPSHOT in the pom.xml is all. That would the latest unstable/snapshot version, it means going towards 0.4.0. Usually the first commit you make on master after making a release branch is to change the version of X to (X+1)-SNAPSHOT.

Often people combine setting the version in the pom and making a release with creating a tag v0.3.0.

lbodor:
But, there will be a few pom files to change, because gws consists of a few maven submodules, maven parent, core, web, and system tests. And all the dependencies in those pom files.

umeat:
Should this specific change be considered a bug fix for version 0.3.0? You could look at it either way, but it doesn't break any interfaces and only modifies the behavior of a yet unused API (which was introduced in this version).

lbodor:
Bug fix, I think

PR #693, opened 2020-05-12
Fix unbound variable errors when running travis/build.sh locally

frankfu03:


lbodor:
Are you trying to test the script locally?

lbodor:
In fact, won't clearing these two variables make the script fail or do the wrong thing when run in travis?

travis/build.sh
```diff
@@ -6,6 +6,9 @@ set -euo pipefail
 # A local installation of maven prefers to run the global installation, if available.
 sudo rm -f /etc/mavenrc
 
+TRAVIS_BRANCH=
+TRAVIS_PULL_REQUEST=
```
lbodor, line 10: What will travis do now?

lbodor: Here is what travis did with this PR: https://travis-ci.org/github/GeoscienceAustralia/Geodesy-Web-Services/builds/686348640

It ran the unit tests, but not the integration tests.

frankfu03: Yes, you are right: it will cause the failure in travis ci. The errors come when I run "`./travis/build.sh`" in my laptop:
`./travis/build.sh: line 13: TRAVIS_BRANCH: unbound variable`

frankfu03: I can manually add the two lines when running locally. So I'll close this PR.

lbodor: You shouldn't run this script locally, it works in combination with .travis.yml. You can just run the bits you want from it locally. We could also at some stage refactor travis/build.sh to invoke a new build.sh script at the root of the repository, which you could run locally,

frankfu03: Good. Thanks.

PR #698, opened 2020-05-15
Allow cross account access to document storage bucket

frankfu03:


aws/terraform/cross_account_roles.tf
```diff
@@ -0,0 +1,48 @@
...
+    {
+      "Action": [
+        "s3:DeleteObject",
+        "s3:PutObject",
+        "s3:PutObjectAcl"
+      ],
+      "Effect": "Allow",
+      "Resource": [
+        "${aws_s3_bucket.gnss_metadata_document_storage.arn}",
+        "${aws_s3_bucket.gnss_metadata_document_storage.arn}/*"
```
lbodor, line 38: You could combine the 2 statement into one.

frankfu03: Yeah, it is done now. 
I thought that it might be clear to have separated read and write policy statements.

frankfu03: I have added a public RSA key to GitHub `ga-gnss-ci` user account and the encrypted private SSH key wrapped in an `openssl` line to before_install stage in the `.travis.yml`.

.travis.yml
```diff
@@ -11,6 +11,11 @@ cache:
   timeout: 86400
   directories:
   - "${HOME}/.m2"
+before_install:
+  - openssl aes-256-cbc -K $encrypted_cc164a3b14b0_key -iv $encrypted_cc164a3b14b0_iv -in github_deploy_key.enc -out github_deploy_key -d
```
lbodor, line 15: I don't see where `-K` and `-iv` values are defined. Should they be listed below as encrypted environment variables?

PR #702, opened 2020-05-21
Add bucket policy to allow cross account access to document storage

frankfu03:
Currently when uploading images to the document storage bucket,  I still got error messages: 
`Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: 27C65B9719E2A5C4)`.
With this bucket policy deployed as testing, I successfully uploaded and deleted images to/from the bucket for ALIC sitelog.

### DIFF
--- a/aws/terraform/document_storage.tf
+++ b/aws/terraform/document_storage.tf
@@ -10,6 +10,33 @@ resource "aws_s3_bucket" "gnss_metadata_document_storage" {
   }
 }
 
+resource "aws_s3_bucket_policy" "document_storage" {
+  bucket = aws_s3_bucket.gnss_metadata_document_storage.id
+
+  policy = <<POLICY
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "s3:DeleteObject",
+        "s3:PutObject",
+        "s3:PutObjectAcl"
+      ],
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "094928090547"
+      },
+      "Resource": [
+        "${aws_s3_bucket.gnss_metadata_document_storage.arn}",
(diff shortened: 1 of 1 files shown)


aws/terraform/document_storage.tf
```diff
@@ -10,6 +10,33 @@ resource "aws_s3_bucket" "gnss_metadata_document_storage" {
...
+      "Resource": [
+        "${aws_s3_bucket.gnss_metadata_document_storage.arn}",
+        "${aws_s3_bucket.gnss_metadata_document_storage.arn}/*"
+      ]
+    }
+  ]
+}
+POLICY
+}
+
```
lbodor, line 39: OK, it's good it works. If you are going with a bucket policy, you should delete the role.

frankfu03: I would like to go with the bucket policy for now. I'll delete the cross account role/policy in another PR.
But I am also want to know how to make the role working. I may have missed something.

lbodor: OK, good. For next time, to make a role work, you have to assume it. Not assuming the role is a bit like making a user, but never logging in as that user.

frankfu03: Thanks. I'll merge it tomorrow morning if you approve it. 
I notice that the Dev GeoServer is stopped at night time everyday. Is it scheduled? Without Dev GeoServer running, the terraform deployment won't run when merging PR.

lbodor: I'm just removed the schedule action to stop the geoserver instance at 8pm. There should be an instance ready in a few minutes.

frankfu03: Good, thanks. 
As we are currently working from home , I sometimes may work late. Otherwise, the scheduled stop actions at night for all instances on dev are good.

PR #704, opened 2020-06-03
Sort EffectiveDates, nulls first

lbodor:


lbodor:
This should fix the ordering of log items in geodesyml sitelogs, and hopefully also in the site manager, unless it imposes a different ordering (I will check). Skeleton files currently only sorts by "from" data. I will send another PR to fix that.

PR #705, opened 2020-06-03
Use latest equipment for skeleton files

lbodor:


gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -71,9 +71,11 @@
             log.error("No receiver record for station " + fourCharId);
             return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
-        receiverLogItemList.sort(Comparator.comparing(GnssReceiverLogItem::getDateInstalled).reversed());
 
-        GnssReceiverLogItem receiverLogItem = receiverLogItemList.get(0);
+        GnssReceiverLogItem receiverLogItem = receiverLogItemList.stream()
+            .max(GnssReceiverLogItem::compareTo)
+            .get();
```
lbodor, line 77: Calling `get()` here is like calling `get(0)`, you get a runtime exception the list is empty,

PR #709, opened 2020-06-11
Remove orphan document objects from s3 bucket for all GNSS sites

frankfu03:


lbodor:
Could this remove documents that have just been uploaded, but not yet attached to site logs?

frankfu03:
Yes, it is likely, but a rare case. On gnss-site-manager, when user clicks "Save" button, all new images will be first uploaded into s3 bucket, and then added to sitelog. If orphan image removal process happens between the steps, then it will happen.

frankfu03:
A possible solution is adding another check of possible orphan images:
1) Find all possible orphan images;
2) Waiting for 5 minutes;
3) Check if any possible orphan images are attached to sitelogs;
4) Remove those that are not attached to any istelogs.

lbodor:
I thought the images were uploaded before the user clicked save.

frankfu03:
It was, but now it has been changed to upload images when user clicks the Save button as per our discussion.

frankfu03:
Based on review, I have made the following changes:

1. An optional parameter variable `hoursToKeep` has been added to the removeOrphanDocuments endpoint. If `hoursToKeep` is not provided (is null), then all orphan documents in s3 bucket will be removed (it is needed for integration tests); if `hoursToKeep` = 2, then orphan documents uploaded within 2 hours will be kept and all other orphans will be removed.

2. Added a shell script for manually running the removeOrphanDocuments endpoint, in which `hoursToKeep` = 24. 

I did some tests. It works properly.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +108,83 @@ public AssociatedDocumentEndpoint() {
...
     }
 
+    @DeleteMapping("/removeOrphanDocuments")
+    public ResponseEntity<String> removeOrphanDocuments(@RequestParam(required = false) Integer hoursToKeep)
+            throws SdkClientException, AmazonServiceException {
+
+        Date startDateTime = null;
+        if (hoursToKeep != null) {
+            DateTime dateTime = new DateTime().minusHours(hoursToKeep);
+            startDateTime = dateTime.toDate();
```
lbodor, line 118: I think that java 8 has incorporated most of functionality of the joda time library. Does your code rely on something in joda time that is unavailable in the standard java library?

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +108,83 @@ public AssociatedDocumentEndpoint() {
...
+        }
+
+        int numberOfOrphanDocuments = 0;
+        List<String> fourCharacterIds = getDistinctFourCharacterIdsInBucket();
+        for (String fourCharacterId : fourCharacterIds) {
+            List<String> documentNames = this.documents.findDocumentNamesByFourCharacterId(fourCharacterId);
```
lbodor, line 124: We should prefer to use `findByFourCharacterId` and then `map` over the result to extract document names. The strategy is to keep class/interface APIs small.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: I didn't check if java 8 includes similar functionality of DateTime from joda library. I'll do it now.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +108,83 @@ public AssociatedDocumentEndpoint() {
...
+                if (!documentNames.contains(objectKey)) {
+                    DeleteObjectRequest deleteObjectRequest = new DeleteObjectRequest(this.bucketName, objectKey);
+                    s3Client.deleteObject(deleteObjectRequest);
+                    numberOfOrphanDocuments++;
+                }
+            }
+        }
+
+        if (numberOfOrphanDocuments == 0) {
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
```
lbodor, line 136: Could we return success here? It feels to me that if a client wants orphans deleted and there aren't any to delete, then that's not an error. It would be different if our api identified a particular document to delete and that document wasn't there.

delete /orphan/1 should return 404 if orphan 1 does't exist
but
delete /orphans should just succeed even if there aren't any to delete.

frankfu03: Good. I'll change to return success.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Ok. I'll change it to `findByFourCharacterId()` which returns a list of Document objects.

gws-core/src/test/java/au/gov/ga/geodesy/support/utils/GMLDateUtilsTest.java
```diff
@@ -90,4 +93,19 @@ public void testStringToDate01() {
...
+        Date dateTarget = simpleDateFormat.parse(dateString);
+
+        String dateString1 = "20200706T001000";
+        Date dateBefore = simpleDateFormat.parse(dateString1);
+        assertThat(dateBefore.compareTo(dateTarget) < 0, is(true));
+
+        String dateString2 = "20200707T003300";
+        Date dateAfter = simpleDateFormat.parse(dateString2);
+        assertThat(dateAfter.compareTo(dateTarget) > 0, is(true));
+    }
```
lbodor, line 110: This test doesn't test any code in src/main/java.

lbodor:
The default value for optional hoursToKeep argument shouldn't be to delete all orphans, because that would delete documents that are about to be associated. The default value should favour production rather than test usage.

gws-core/src/test/java/au/gov/ga/geodesy/support/utils/GMLDateUtilsTest.java
frankfu03: The code has been changed a lot and this unit test is not relevant now. I'll removed it. 
Previously I checked the document's dateTaken values for hourToKeep, and now I changed to check the last modified time.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Yes, the joda time library has been included in Java 8 with the java.time package. I changed to use LocalDateTime which has similar functionality as joda's DateTime class.

lbodor: LocalDateTime may not be the right choice. It doesn't correspond the a point on the timeline, because it doesn't not specify a timezone. Your use-case is for a point on the timeline. Time is hard, an unexpected timezone can be applied and affect your comparison. Hopefully, the java.time.* API won't let you make a mistake, i.e., you may be required to specify a timezone at some stage.

lbodor: Consider using Instant, you can say Instant.now(), minus(), and you can go back to java.util.Date with, Date.from.

frankfu03:
Considering that the default value should favour production, I can make the changes: if there is no `hoursToKeep` in the request URL, then set `hoursToKeep` to a specific value, like 12; set `hoursToKeep=0` for the AssociateDocumentEndpointITest. 

Or even simple, change `hoursToKeep` to be a required parameter.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
lbodor: Both java.time.Instant and java.util.Date values are relative to the unix epoch.

lbodor:
Yes, that's a good idea.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: As the Document class does not have a field `FourCharacterId`, both methods `List<String> findDocumentNamesByFourCharacterId()` and `List<Document> findByFourCharacterId()` need the `@Query("select d.name _or d_ from Document d where d.name like concat(UPPER(:fourCharacterId),'%')")` annotation in the interface. I prefer to no change here as it has less efforts.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Thanks for the info. I changed to use Instant.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +109,83 @@ public AssociatedDocumentEndpoint() {
...
+                    numberOfOrphanDocuments++;
+                }
+            }
+        }
+
+        if (numberOfOrphanDocuments == 0) {
+            return new ResponseEntity<String>(HttpStatus.OK);
+        } else {
+            log.info(numberOfOrphanDocuments + " orphan documents have been removed from " + this.bucketName);
+            return new ResponseEntity<String>(HttpStatus.NO_CONTENT);
```
lbodor, line 140: Is it important to send the client different success codes? I think that the advice people give regarding NO_CONTENT and DELETE applies when the client requests to delete a single resource, which isn't the case here. If would be simpler to always return OK and a body saying how many documents where deleted. It's ok for the server-side log, but there is no need to mention the bucket to the client.

lbodor: You would not need the if-statement anymore.

frankfu03: The if-else statement has been changed to
`return ResponseEntity.ok().body("Number of orphan documents removed: " + numberOfOrphanDocuments);`

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +111,80 @@ public AssociatedDocumentEndpoint() {
...
+            Instant dateTime = Instant.now().minus(hoursToKeep, ChronoUnit.HOURS);
+            startDateTime = Date.from(dateTime);
+        }
+
+        int numberOfOrphanDocuments = 0;
+        List<String> fourCharacterIds = getDistinctFourCharacterIdsInBucket();
+        for (String fourCharacterId : fourCharacterIds) {
+            List<Document> documents = this.documentRepository.findByFourCharacterId(fourCharacterId);
+            List<String> documentNames = documents.stream().map(d -> d.getName()).collect(Collectors.toList());
+            List<String> objectKeys = this.getObjectKeys(fourCharacterId, startDateTime);
```
lbodor, line 129: Will this fetch the same object keys from AWS S3 twice, once on line 125 and again on line 129?

lbodor: What if you went to AWS s3 once to return `HashSet<List, List<String>>`, where the key is the four character id and the value is the list of keys beginning with that four character id. You wouldn't have to care about distinct four char ids, because the set will take care of that four you, and you won't have to go to s3 twice, because you will already have the keys.

The way you have it, you go to s3 once to get the keys, you extract four char ids and throw away the keys, then you go s3 again to key the keys again.

lbodor: Do you need to group object keys and document names by four character id? Would it work to just find all object keys and all find documents, filter, and delete?

frankfu03: I have refactored the code: it first finds all object keys and all documents, and then delete orphan objects from the bucket.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
```diff
@@ -14,4 +15,7 @@
     List<Document> findAll();
 
     Document findByName(@Param("name") String name);
+
+    @Query("select d from Document d where d.name like concat(UPPER(:fourCharacterId),'%')")
+    List<Document> findByFourCharacterId(@Param("fourCharacterId") String fourCharacterId);
```
lbodor, line 20: It looks like this isn't used anymore.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +111,61 @@ public AssociatedDocumentEndpoint() {
...
+        List<String> documentNames = documents.stream().map(d -> d.getName()).collect(Collectors.toList());
+
+        int numberOfOrphanDocuments = 0;
+        for (String objectKey : objectKeys) {
+            if (!documentNames.contains(objectKey)) {
+                DeleteObjectRequest deleteObjectRequest = new DeleteObjectRequest(this.bucketName, objectKey);
+                s3Client.deleteObject(deleteObjectRequest);
+                numberOfOrphanDocuments++;
+            }
+        }
```
lbodor, line 135: It is really easy to see what's going on now.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
lbodor: It looks good, thank you.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
frankfu03: Yes, it has been removed from the interface.

PR #711, opened 2020-06-11
Create a http get endpoint to retrieve the content of a document

frankfu03:


gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -59,6 +65,18 @@ public AssociatedDocumentEndpoint() {
...
 
+    @GetMapping(
+        value = "/{name}"
+    )
+    public ResponseEntity<String> getDocumentContent(@PathVariable("name") String documentName)
+            throws SdkClientException, AmazonServiceException, IOException {
+        GetObjectRequest getObjectRequest = new GetObjectRequest(this.bucketName, documentName);
+        S3Object documentObject = this.s3Client.getObject(getObjectRequest);
+        InputStream inputStream = documentObject.getObjectContent();
+        String documentContent = StreamUtils.copyToString(inputStream, StandardCharsets.UTF_8);
```
lbodor, line 76: Images are binary, so they shouldn't be decoded as UTF-8, which is text. Minimum change would be to return `ResponseEntity<byte[]>`, but it would be even better not to load the image into memory, but to stream it to the client. Take a look at Spring MVC options for streaming data from an endpoint.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpointITest.java
```diff
@@ -13,20 +15,21 @@
...
     private String fileReference = "http://localhost:4572/gnss-metadata-document-storage-local/" + this.documentName;
+    private String documentContent = "Dummy testing image content";
 
     @Test
     @Rollback(false)
     public void uploadDocument() throws Exception {
-        InputStream fileContent = new ByteArrayInputStream("Dummy testing image content".getBytes());
-        MockMultipartFile mockFile = new MockMultipartFile("file", documentName, "image/jpg", fileContent);
+        InputStream fileContent = new ByteArrayInputStream(this.documentContent.getBytes());
+        MockMultipartFile mockFile = new MockMultipartFile("file", this.documentName, "image/jpg", fileContent);
```
lbodor, line 32: This test works, because the image content is a UTF-8, but in reality that won't be the case. I think that document content should be streamed to the client uninterpretted.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
frankfu03: Yes, I used bytes[] before, but could not get it pass the integration test which checks its content. So I would like to deploy and test it via gws dev.

gws-core/src/test/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpointITest.java
frankfu03: I'll try streaming data from endpoint to client later.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
lbodor: Feel free to do that. But, you could also test it by deploying gws in docker using our docker-compose setup in this project. It should give you the same result as running gws in dev. Also, I suspect that no matter what outcome you get from the test in gws dev, we will still want to treat these documents as binary streams, rather than text.

frankfu03:
This PR works fine, but it could not override the default cache settings (no cache, no store) which was added by Spring Security.  This issue may be resolved in latest Spring release.

PR #712, opened 2020-06-18
Add signal types to skeleton files

umeat:
I think the better way to do this would be to add parsing of signal types into the support-rinex header class. That would definitely take me longer than this took me and we would still have the mapping of signal types to receiver type in the skeleton class.

Any thoughts on a better way to insert the signals into the header string? Maybe `len(headerString) - 80` (or 81 or whatever) would be nicer to get the lines in before the "END OF HEADER" line. Or splitting the string on \n and inserting in the second last element (which I looked into but it involved a lot more code than I expected).

### DIFF
--- a/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
+++ b/gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
@@ -19,7 +19,6 @@ import org.springframework.web.bind.annotation.*;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.List;
 
 @RestController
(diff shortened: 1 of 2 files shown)


umeat:
https://gajira.atlassian.net/secure/RapidBoard.jspa?rapidView=263&projectKey=GAGA&modal=detail&selectedIssue=GAGA-434

lbodor:
Yep, it should probably go into support-rinex. There may be better ways to choose the right spot, but if this is temporary, then it doesn't matter, once we move it RinexFileHeader class we should have a good way of selecting the right line.

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/SkeletonEndpoint.java
```diff
@@ -104,9 +124,18 @@
 
         StringWriter stringWriter = new StringWriter();
         rinexFileHeader.write(stringWriter);
+        String headerString = stringWriter.toString();
+
+        // Assume signal types based on receiver type - unknown receiver types will be ignored
+        String receiverType = receiverLogItem.getType().toLowerCase();
+        if (receiverType.contains("leic") || receiverType.contains("sept")) {
+            headerString = headerString.substring(0, 538) + septLeicaObsTypes + headerString.substring(538);
```
lbodor, line 132: But unless you know that 538 is always right, counting from the back to skip end-of-header line would be better.

umeat: There are no continuation lines in the header, and this process doesn't add comments so we know that 538 is always correct. But might be clearer to count backwards and comment that it's being inserted before the END OF HEADER line (which always has to be the last line, so the length is definitely consistent).

lbodor: There is also `new StringBuilder(str).insert(538, str2).toString()`

umeat:
Should I bump the version in the pom for this change?

lbodor:
I'm not sure that it matters. We can just keep rolling with the snapshot release. This project is not a dependency of any other project.

PR #715, opened 2020-06-19
Create a http get endpoint for redirecting to s3 object urls

frankfu03:


lbodor:
Generally, web applications don't need to be told their address. The servlet http request object should have enough information in to construct an absolute URL back into GWS.

frankfu03:
The dev geodesy-site-manager (https://dev.gnss-site-manager.geodesy.ga.gov.au/) works fine with the redirect url for site images. 
But the http get endpoint for a site image (for example: `https://devgeodesy-webservices.geodesy.ga.gov.au/associatedDocuments/ADE1_ant_090_20200623T235048.jpg`) will show the redirected URL:
`https://s3-ap-southeast-2.amazonaws.com/gnss-metadata-document-storage-dev/ADE1_ant_000_20200623T235048.jpg`

gws-core/src/main/java/au/gov/ga/geodesy/port/adapter/rest/AssociatedDocumentEndpoint.java
```diff
@@ -95,12 +102,27 @@ public AssociatedDocumentEndpoint() {
...
         log.error(e.getMessage(), e);
         return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(e.getMessage());
     }
 
+    private String getDocumentLocation(HttpServletRequest request, String documentName) {
+        String documentLocation = request.getRequestURL() + documentName;
+        if (request.getScheme().equals("http") && !request.getServerName().contains("localhost")) {
+            documentLocation = documentLocation.replace("http", "https");
+        }
+        return documentLocation;
```
lbodor, line 123: Could we return a relative location instead?

frankfu03: I remember I tried to return a relative location here, but got a problem in site-manager. The returned url is incorrect.

lbodor: The returned url would relative and the site manager would need to make it absolute. How was it incorrect?

frankfu03: It is much easy to return an absolute url here.
If we return a relative location, this relative location will be saved to `fileReference` in sitelog geodesyml file. We can not directly view the site image via the `fileReference` in web browser. Although the relative location can be made absolute in site manager, it takes extra effort. Besides there are external urls.

lbodor: We return a relative URL to the client, the client makes the URL absolute by prepending the hostname and inserts the absolute URL into the sitelog.

It's not easy to return an absolute URL, because the server generally does not know it's URL, but the client does. The client has contacted the server, so it knows the exact URL. The code above is also extra effort, and it attempts to work around the fact that the server does not know it's URL. For example, in the case above, how does the server know that it's "http://localhost" and not "https://localhost"? It's just a guess. The decision to encrypt traffic is made by the AWS ELB, and the server is unaware of that fact. Other mappings/transformations of the URL are possible that would also have no bearing on what `request.getRequestURL` returns.

frankfu03: Good. I have changed to return a relative URL.

PR #717, opened 2020-06-29
Create a Document table and a joint table to replace AssociatedDocument table

frankfu03:
Changes are based on https://gajira.atlassian.net/browse/GNSSSM-100.


gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
```diff
@@ -0,0 +1,23 @@
...
+public interface DocumentRepository extends
+        DocumentRepositoryCustom,
+        JpaRepository<Document, Integer>,
+        QueryDslPredicateExecutor<Document> {
+
+    @Override
+    List<Document> findAll();
+
+    @Query("select d from Document d where d.id = :id")
+    Document findById(@Param("id") Integer id);
```
lbodor, line 19: Do you need this method? Doesn't find by id exist always?

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Query("select d from Document d where d.name = :name")
```
lbodor, line 21: I don't think you need this query. Spring data rest should be able to generate it from the method name. If you have a method on this interface called findByName, then the framework will generate an implementation, by looking at the method name.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepositoryCustom.java
```diff
@@ -1,6 +1,6 @@
 package au.gov.ga.geodesy.domain.model.sitelog;
 
-public interface AssociatedDocumentRepositoryCustom {
+public interface DocumentRepositoryCustom {
     void delete(Integer id);
     void delete(String name);
```
lbodor, line 5: It would be simpler to first call findByName and then delete by id, which should already exist. We could then omit this custom interface altogether.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepositoryImpl.java
```diff
@@ -0,0 +1,23 @@
...
+        }
+    }
+
+    public void delete(String name) {
+        Document toDelete = documents.findByName(name);
+        if (toDelete != null) {
+            documents.delete(toDelete);
+        }
+    }
+}
```
lbodor, line 23: We could also remove this implementation. Delete by id already exists. Delete by name we can archive by calling find by name and then delete.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
frankfu03: Not used yet. Do you mean that findById() exists by default?

gws-core/src/main/resources/db/add-document-table.xml
```diff
@@ -0,0 +1,44 @@
...
+                             tableName="document"
+                             constraintName="uk_document_name"
+                             deferrable="true"
+                             initiallyDeferred="true"/>
+        <addUniqueConstraint columnNames="file_reference"
+                             tableName="document"
+                             constraintName="uk_document_filereference"
+                             deferrable="true"
+                             initiallyDeferred="true"/>
+        <setTableRemarks remarks="table contains documents (site map, site diagram, site images, etc.) in site log"
```
lbodor, line 40: not site log, could be anything

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
lbodor: Yes. Here it is: https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
frankfu03: I'll test it by removing @Query annotation.
There are a few repository classes do not have the @Query annotation, but more have, eg., `public List<Node> findBySiteId(@Param("siteId") Integer siteId)` in NodeRepository, `SiteLog findByFourCharacterId(@Param("id") String id)` in SiteLogRepository.

gws-core/src/main/resources/db/add-document-table.xml
frankfu03: I'll remove the remarks line.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepositoryImpl.java
frankfu03: Ok. I'll remove this implementation.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepositoryCustom.java
frankfu03: Good. I'll remove this interface.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
frankfu03: Yeah, removed this method.

gws-core/src/main/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepository.java
frankfu03: The @query annotation has been removed.

frankfu03:
All issues have been addressed and changes were made accordingly. It is ready for review again.

gws-core/src/test/java/au/gov/ga/geodesy/domain/model/sitelog/DocumentRepositoryITest.java
```diff
@@ -11,47 +8,49 @@
...
-        AssociatedDocument d1 = documents.iterator().next();
+        Document d1 = documents.iterator().next();
         assertThat(d1.getName(), is("MOBS_ant_000_20191027T143000.jpg"));
 
-        associatedDocuments.saveAndFlush(d1);
-        AssociatedDocument d2 = associatedDocuments.findByName(d1.getName());
+        documentRepository.saveAndFlush(d1);
+        Document d2 = documentRepository.findByName(d1.getName());
         assertThat(d2.getName(), is(d1.getName()));
     }
```
lbodor, line 55: Are method names `saveMobs` and `saveAlice` copied from some other test? The tests in this class don't save sitelogs for mobs and alic. They save the documents, so would just one test method be sufficient. How are they different?

frankfu03: Yes, both test methods are copied from other testing classes. There are no much difference: one saves two documents for ALIC and another saves one for MOBS;  one checks a name in an array, and the other is exact name matching.

I can do the changes in a separate PR: rename saveAlic() to saveDocumentsForAlic(), and delete saveMobs() .

PR #722, opened 2020-07-31
Fill in mandatory fields for sitelog geodesyml files used in e2e tests

frankfu03:


gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade1_20080530.xml
```diff
@@ -131,13 +131,13 @@
                 <geo:tiedMarkerCDPNumber>(A4)</geo:tiedMarkerCDPNumber>
                 <geo:tiedMarkerDOMESNumber>(A9)</geo:tiedMarkerDOMESNumber>
                 <geo:differentialComponentsGNSSMarkerToTiedMonumentITRS>
-                    <geo:dx xsi:nil="true"></geo:dx>
-                    <geo:dy xsi:nil="true"></geo:dy>
-                    <geo:dz xsi:nil="true"></geo:dz>
+                    <geo:dx xsi:nil="true">0.0</geo:dx>
+                    <geo:dy xsi:nil="true">0.0</geo:dy>
+                    <geo:dz xsi:nil="true">0.0</geo:dz>
```
lbodor, line 136: Are these mandatory?

See https://groups.google.com/forum/#!topic/geodesyml/oZn8K_vWohM

gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade1_20080530.xml
```diff
@@ -215,14 +215,14 @@
         </geo:temperatureSensor>
         <geo:waterVaporSensor>
             <geo:WaterVaporSensor gml:id="water-vapor-1">
-                <geo:type codeSpace="urn:ga-gov-au:water-vapor-sensor-type">UNK</geo:type>
-                <geo:manufacturer></geo:manufacturer>
-                <geo:serialNumber></geo:serialNumber>
+                <geo:type codeSpace="urn:ga-gov-au:water-vapor-sensor-type">WVSS-II</geo:type>
```
lbodor, line 218: Where does the value WVSS-II come from?

gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade2_20090901.xml
```diff
@@ -140,8 +140,8 @@
                 <geo:igsModelCode codeList="http://xml.gov.au/icsm/geodesyml/codelists/antenna-receiver-codelists.xml#GeodesyML_GNSSAntennaTypeCode" codeListValue="ASH700936B_M SNOW" codeSpace="urn:xml-gov-au:icsm:egeodesy:0.5">ASH700936B_M    SNOW</geo:igsModelCode>
                 <geo:antennaReferencePoint codeSpace="urn:ga-gov-au:antenna-reference-point-type">BPA</geo:antennaReferencePoint>
                 <geo:marker-arpUpEcc.>0.0</geo:marker-arpUpEcc.>
-                <geo:marker-arpNorthEcc. xsi:nil="true"></geo:marker-arpNorthEcc.>
-                <geo:marker-arpEastEcc. xsi:nil="true"></geo:marker-arpEastEcc.>
+                <geo:marker-arpNorthEcc.>0.0</geo:marker-arpNorthEcc.>
+                <geo:marker-arpEastEcc.>0.0</geo:marker-arpEastEcc.>
```
lbodor, line 144: These are nillable in the schema.

gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade2_20090901.xml
```diff
@@ -159,13 +159,13 @@
                 <geo:tiedMarkerCDPNumber>(A4)</geo:tiedMarkerCDPNumber>
                 <geo:tiedMarkerDOMESNumber>(A9)</geo:tiedMarkerDOMESNumber>
                 <geo:differentialComponentsGNSSMarkerToTiedMonumentITRS>
-                    <geo:dx xsi:nil="true"></geo:dx>
-                    <geo:dy xsi:nil="true"></geo:dy>
-                    <geo:dz xsi:nil="true"></geo:dz>
+                    <geo:dx xsi:nil="true">0.0</geo:dx>
+                    <geo:dy xsi:nil="true">0.0</geo:dy>
+                    <geo:dz xsi:nil="true">0.0</geo:dz>
```
lbodor, line 164: I've checked the schema here and geo:differentialComponentsGNSSMarkerToTiedMonumentITRS is nillable, so
```
<geo:differentialComponentsGNSSMarkerToTiedMonumentITRS xsi:nil="true"/>
```
Should be OK.

lbodor:
Working out what's mandatory and what's not is a bit tricky. We need GeodesyML domain modelling expertise to work this out. Do the tests break without these changes? We also want to be able to save invalid data.

gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade1_20080530.xml
frankfu03: It is a make-up value.

gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade1_20080530.xml
frankfu03: Good, I'll remove the changes for the three fields.

gws-system-test/src/main/resources/sitelogs/2017-06-29/geodesyml/ade2_20090901.xml
frankfu03: Yes, you are right. But if keeping `xsi:nil="true"`, these two fields will be empty.

frankfu03:
Sometimes e2e tests failed with unexpected/unknown errors. I noticed that some of the failures were related to the input sitelog geodesyml. Although my intention is to fill in values for mandatory fields, I think it would be good to fill most optional fields as well, only leaving a few fields empty for testing purposes.

lbodor:
I remember this coming up before in a different scenario, but still quite similar. E2E tests of the site manager UI aren't system tests and shouldn't be touching the deployed instance of GWS in dev. In e2e tests, we need to mock calls to the GWS to respond with pre-canned site log files, which are stored in the site manager repository.

frankfu03:
Isn't it that e2e tests for site manager use sitelog geodesyml from dev GWS?

lbodor:
They do, but they shouldn't The e2e tests should not touch GWS in dev. E2E tests are integration tests, where end-to-end stands for all components in the site manager application, not end-to-end of our deployed applications/systems. Unit tests target individual components, integration tests target all components together, but still within an application, before it is deployed. If we had system tests for the site manager, they would run against a deployed site manage instance (https://dev-gnss-site-manager.geodesy.ga.gov.au), which then would use GWS in dev.

E2E integration tests run against each pull request, so they can run concurrently, so they should run in isolation. But if they are touching a shared deployed resources (GWS in dev) then there will be interference.

frankfu03:
I agree that we should use mock calls to GWS to respond with pre-canned site log files. It is an ideal scenario for e2e tests. But we have to figure out how to save a sitelog with mock calls.

frankfu03:
Not relevant anymore

PR #725, opened 2020-08-19
Upload and save associated documents using xsltproc and XSLT stylesheet

frankfu03:


gws-system-test/src/main/bash/add-associated-document.sh
```diff
@@ -0,0 +1,118 @@
...
+    curl -X POST -F "file=@${documentPath}" \
+                 -F "documentName=${documentName}" \
+                 -H "Authorization: Bearer ${jwt}" \
+                 "${gws}/associatedDocuments/"
+
+    outputSiteLogXml=$(xsltproc --stringparam description "$description" \
+                                --stringparam documentName "$documentName" \
+                                --stringparam contentType "$contentType" \
+                                --stringparam createdDate "$createdDateFmt" \
+                                --stringparam fileReference "$fileReference" \
```
lbodor, line 112: Why do we need to supply document name when uploading via this script and not when uploading using the site manager?

gws-system-test/src/main/bash/add-associated-document.sh
```diff
@@ -0,0 +1,118 @@
...
+        description="Antenna Monument"
+    elif [[ $1 == "ant_bldg" ]]; then
+        description="Antenna Building"
+    elif [[ $1 == "ant_roof" ]]; then
+        description="Antenna Roof"
+    else
+        echo "Unknown document code"
+        echo "Valid document code: ant_000, ant_090, ant_180, ant_270, ant_sn, rec_sn, ant_monu, ant_bldg, ant_roof"
+        usage
+    fi
```
lbodor, line 42: Since you're always comparing against $1 (a single value), then a case statement may be more appropriate.

gws-system-test/src/main/bash/add-associated-document.sh
```diff
@@ -0,0 +1,118 @@
...
+              "${openam}/oauth2/access_token?realm=/" | jq -r .id_token)
+
+inputSiteLogXml=/tmp/sitelog_input.xml
+curl "${gws}/siteLogs/search/findByFourCharacterId?id=${fourCharacterId}&format=geodesyml" > ${inputSiteLogXml}
+
+if [[ -f ${inputSiteLogXml} ]]; then
+    documentName="${fourCharacterId}_${documentCode}_${createdDate}.${documentPath##*.}"
+    contentType=$(file --mime-type -b "$documentPath")
+    createdDateFmt="${createdDate:0:4}-${createdDate:4:2}-${createdDate:6:5}:${createdDate:11:2}:${createdDate:13:2}.000Z"
+    fileReference="${gws}/associatedDocuments/${documentName}"
```
lbodor, line 101: Quotation is a bit inconsistent. I don't think you ever need a quote immediately to the right of `=`.

```
documentName=${fourCharacterId}_${documentCode}_${createdDate}.${documentPath##*.}
```
but
```
contentType=$(file --mime-type -b "$documentPath")
```

gws-system-test/src/main/bash/add-associated-document.sh
lbodor, line 112: Use of curly braces is also a bit inconsistent. I tend to not used, unless they are needed, and generally they aren't needed.

gws-system-test/src/main/bash/add-associated-document.sh
```diff
@@ -0,0 +1,118 @@
...
+clientPassword=
+username=
+password=
+
+jwt=$(curl -s --user ${clientId}:${clientPassword} \
+              --data "grant_type=password&username=${username}&password=${password}&scope=openid profile" \
+              "${openam}/oauth2/access_token?realm=/" | jq -r .id_token)
+
+inputSiteLogXml=/tmp/sitelog_input.xml
+curl "${gws}/siteLogs/search/findByFourCharacterId?id=${fourCharacterId}&format=geodesyml" > ${inputSiteLogXml}
```
lbodor, line 95: If xsltproc can read from the standard input, then you won't need the temporary file.

gws-system-test/src/main/bash/add-associated-document.sh
```diff
@@ -0,0 +1,118 @@
...
+elif [[ $# -lt 3 ]]; then
+    echo "Error: Four Character Id is empty."
+    usage
+elif [[ $# -lt 4 ]]; then
+    echo "Error: document code is empty."
+    usage
+elif [[ $# -lt 5 ]]; then
+    echo "Error: document created date is empty."
+    usage
+elif [[ ! $5 =~ ^[0-2][0-9]{3}(0[1-9]|1[012])(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3])([0-5][0-9]){2}$ ]]; then
```
lbodor, line 66: Will the API reject a data if not in correct format? If yes, then it would be simpler to let the API reject invalid dates.

gws-system-test/src/main/bash/add-associated-document.sh
```diff
@@ -0,0 +1,118 @@
...
+}
+
+if [[ $# -lt 1 ]]; then
+    echo "Error: environment is empty."
+    usage
+elif [[ $1 != "dev" && $1 != "test" && $1 != "prod" ]]; then
+    echo "Error: invalid environment."
+    usage
+elif [[ $# -lt 2 ]]; then
+    echo "Error: document path is empty."
```
lbodor, line 52: There is probably a way of leveraging `set -u`, which is turned on at the top, to avoid testing for empty variables.

gws-system-test/src/main/bash/add-associated-document.sh
lbodor, line 101: Why not use the file reference as returned in the location response header?

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Both are supplied with the document name. In Java and Typescript/Javascript, the document name is added to MultipartFile as originalFilename. For example,

- In AssociatedDocumentEndpoint:

```
    public ResponseEntity<String> upload(@RequestPart("file") MultipartFile file)
            throws SdkClientException, AmazonServiceException, URISyntaxException, IOException {
        String documentName = file.getOriginalFilename();
```

-  in AssociatedDocumentEndpointITest:

    `MockMultipartFile mockFile = new MockMultipartFile("file", documentName, "image/jpg", fileContent);`

- In GNSS-Site-Manager:

```
let data = new FormData();
        data.set('file', file, documentName);
        return this.http.post(this.resourceUrl, data, {headers: this.getHttpHeaders()})
```

While in shell script, I couldn't figure out how to add document name to the form param `-F "file=@$documentPath"`.  So I just added another form param `-F "documentName=$documentName"`, and added optional param `@RequestParam(required = false) String documentName` to AssociatedDocumentEndpoint `upload()` method.

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Yes, I changed it to a case statement.

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Quotations have been removed.

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Removed.

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Yes, xsltproc will use the standard input if a filename provided is -.

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Yes, the API rejects invalid date:
`java.lang.RuntimeException: Failed to parse 20191212T121212 using the following patterns: uuuu-MM-dd&#39;T&#39;HH:mm:ss.SSSXXX, uuuu-MM-dd&#39;T&#39;HH:mm:ss.SSSZ, uuuu-MM-dd&#39;T&#39;HH:mm:ssX, uuuu-MM-ddX, uuuu-MM-dd, uuuu-MM, uuuu`
The `createdDate `is a field in Document table and used to compose the document name. Would it be good if the input params are checked at the beginning of the script? The usage function shows the supported format of date input.

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Yes, it is done.

gws-system-test/src/main/bash/add-associated-document.sh
lbodor: I think it would be better for the API to decide what is valid to avoid maintaining the rules in two places. Since this isn't a user facing program, unlike the site manager, the error message from API is good enough.

You mention that the date is used in the document name. But if it's invalid, then it won't be used because the API will reject it?

lbodor: Also, does the above error message indicate that the API will accept a date in a format that's a little easier to read and write, like 2002-12-12T02:03:00?

frankfu03: The document name is in format of ${fourCharacterId}_${documentCode}_$**createdDate**.$extension. For example, ALIC_ant_000_20200820T140000.jpg. It will be used even if it is invalid. 
The **createdDate** is a field in Document table. It must be a format of uuuu-MM-dd, uuuu-MM-ddTHH:mm:ss.SSSZ, etc. The API will reject it if createdDate is invalid.

lbodor: OK, got it, thanks.

lbodor (approved):

gws-system-test/src/main/bash/add-associated-document.sh
frankfu03: Just tested: 2002-12-12T02:03:00 is rejected by API.
It only supports the following patterns: uuuu-MM-ddTHH:mm:ss.SSSXXX, uuuu-MM-ddTHH:mm:ss.SSSZ, uuuu-MM-ddTHH:mm:ssX, uuuu-MM-ddX, uuuu-MM-dd, uuuu-MM, uuuu.

frankfu03: The above line in the script is to ensure createdDate is in format of `uuuuMMddTHHmmss`. An example of command line to run this script:
`gws-system-test/src/main/bash/add-associated-document.sh dev ~/Downloads/ALIC00AUS-20170810-N.jpg ALIC ant_270 20191212T105858`
Where, `20191212T105858` is the createdDate and is used to compose the documentName: `ALIC_ant_270_20191212T105858.jpg` which is the s3 object name and the `name` field in Document table.  `20191212T105858` is then reformatted to `2019-12-12T10:58:58.000Z` so that it can be inserted into the Document table via the API. 
In GNSS-Site-Manager, the createdDate is also formatted to `uuuu-MM-ddTHH:mm:ss.SSSZ` before saving site images to sitelog via the API.

PR #729, opened 2020-10-20
Restore setTableRemarks

lbodor:
This change set was already applied to the test database, so it can't be
modified.


lbodor:
There is no need to add another change set to remote the table remarks, because the next changset file, will remote the entire table.

